PR #49, opened 2015-11-16
[FEATURE] Entry/EntryGroup

sentientwaffle:
Requires https://github.com/interledger/five-bells-shared/pull/35.

Also: https://github.com/interledger/passport-http-signature will need to be published for this to work.

Once it is, npm-shrinkwrap will need to be updated.


justmoon:
LGTM. Found a minor issue (#51) which we can fix later.

PR #69, opened 2015-12-09
[FEATURE] Require auth for GET /accounts/*

### LINKED ISSUE #67: Require auth for GET /accounts/:id
GET /accounts/:id should require auth either from the user (`:id`) or from an admin. I.e. you can see only your own balance unless you're an admin. Then you can see anyone's balance.

sentientwaffle:
Fixes #67

### DIFF
--- a/app.js
+++ b/app.js
@@ -45,8 +45,13 @@ router.put('/transfers/:id',
 router.get('/transfers/:id', transfers.getResource)
 router.get('/transfers/:id/state', transfers.getStateResource)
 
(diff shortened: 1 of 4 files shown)


justmoon:
I think I prefer 403 - you know how Github always confuses everybody with their 404s when you don't have access?

The message we're trying to give the user is: You don't have access to this and you're not supposed to know if this even exists. Since that is a permission issue first, 403 seems appropriate.

LGTM

PR #72, opened 2015-12-17
Add ability to delete accounts

justmoon:
Adds a new endpoint `DELETE /accounts/:id` for deleting existing accounts. Only the administrator is allowed to delete an account.

*Not* ready to merge! The current implementation will not allow deleting accounts that have transfers associated with them (foreign key constraints). It seems unwise to allow the deletion of transfers - the ledger history should never be altered. So perhaps instead of deleting accounts, what we actually want is a way to disable accounts.

cc/ @MatthewPhinney

### DIFF
--- a/app.js
+++ b/app.js
@@ -57,6 +57,10 @@ router.put('/accounts/:id',
   filterAdmin,
   models.Account.createBodyParser(),
   accounts.putResource)
+router.delete('/accounts/:id',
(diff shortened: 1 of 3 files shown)


MatthewPhinney:
Makes sense. Perhaps we can just add a `disabled` boolean to the `Accounts` table in the db ... or something along those lines.

justmoon:
Closing (won't fix) in favor of #75.

PR #74, opened 2015-12-18
[FIX] Disallow transfer to/from same account

MatthewPhinney:


MatthewPhinney:
Disallow transfer requests in which the same account appears in both the credits and debits lists. There are a few ways to handle this.  One would be to add a check to the JSON schema validation, but given the way the code is currently written, I thought it made more sense to put the logic in `transfers.js`. That being said, I'm open to suggestions.

justmoon:
Why would this be disallowed?

I can see a number of cases where it may be nice to have this capability. Let's say there is an API that's listening for payments and triggering actions based on them. Why shouldn't you be able to trigger those actions yourself?

Another example would be a smart contact where you have to pay out to three people equally. Why shouldn't the payor be allowed to be one of those people?

Yes theoretically moving money out of an account and in again is a no-op, but allowing that no-op is totally safe and much easier than to special case around it.

MatthewPhinney:
@vhpoet did you have a particular use case in mind when you opened this issue?

vhpoet:
Yea, my issue was when there's a transfer from a single entity to a single entity. In that case I don't know of any use cases where it makes sense, but that might be my poor knowledge of how the ledger is designed. Anyway I agree with @justmoon that it's not technically a problem, so maybe no point in adding a special case. The check probably makes more sense to do on the client side.

justmoon:
Closing. Won't fix.

PR #76, opened 2015-12-24
Remove pre_* states

justmoon:
The `pre_prepared` and `pre_executed` states were intended to provide a way to handle cases where `five-bells-ledger` has to wait for some external system to prepare or execute a transfer.

However, in practice these states expose implementation details of the ledger which a client cannot usefully react to. A transfer is either `prepared` or it isn't. There is no need for a separate "I'm trying to prepare this transfer" state.

Therefore, `pre_prepared` and `pre_executed` should be removed to simplify the implementation.


emschwartz:
LGTM

PR #78, opened 2015-12-24
Test against three databases

### LINKED ISSUE #64: CircleCI tests should test against MySQL, Postgres and SQLite
We keep running into errors that only appear with certain databases. Let's improve the CI tests to run the whole suite on these three types of database:

* MySQL
* Postgres
* SQLite

justmoon:
Closes #64.

Run test suite against MySQL, Postgres and SQLite.


emschwartz:
LGTM

PR #84, opened 2016-01-04
[FEATURE] /connectors and always 200 for transfer state

sentientwaffle:
* Add unauthenticated `/connectors` endpoint.
* `/state`: sign nonexistent transfers; allow specification of `signatureType` and state.
* Requires https://github.com/interledger/five-bells-shared/pull/42

### DIFF
--- a/migrations/20150820161813-init.js
+++ b/migrations/20150820161813-init.js
@@ -6,7 +6,7 @@ module.exports = {
       primary: { type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true },
       name: { type: Sequelize.STRING, unique: true },
       balance: Sequelize.DECIMAL(10, 2),
-      identity: Sequelize.STRING(1024),
+      connector: Sequelize.STRING(1024),
       password: Sequelize.STRING,
       public_key: Sequelize.TEXT,
       is_admin: Sequelize.BOOLEAN,
--- a/migrations/20150820161813-init.js
+++ b/migrations/20150820161813-init.js
@@ -6,7 +6,7 @@ module.exports = {
       primary: { type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true },
       name: { type: Sequelize.STRING, unique: true },
       balance: Sequelize.DECIMAL(10, 2),
-      connector: Sequelize.STRING(1024),
+      identity: Sequelize.STRING(1024),
       password: Sequelize.STRING,
       public_key: Sequelize.TEXT,
       is_admin: Sequelize.BOOLEAN,
--- a/src/controllers/accounts.js
+++ b/src/controllers/accounts.js
@@ -14,6 +14,13 @@ exports.getCollection = function * find () {
   this.body = _.invoke(accounts, 'getDataExternal')
 }
 
+exports.getConnectors = function * () {
+  const accounts = yield Account.findAll({
+    where: { identity: { $ne: null } }
+  })
+  this.body = _.invoke(accounts, 'getDataConnector')
+}
+
 /**
  * @api {get} /accounts/:id Fetch user info
  * @apiName GetAccount
--- a/src/controllers/accounts.js
+++ b/src/controllers/accounts.js
@@ -16,7 +16,7 @@ exports.getCollection = function * find () {
 
 exports.getConnectors = function * () {
   const accounts = yield Account.findAll({
-    where: { identity: { $ne: null } }
+    where: { connector: { $ne: null } }
   })
   this.body = _.invoke(accounts, 'getDataConnector')
 }
--- a/src/controllers/transfers.js
+++ b/src/controllers/transfers.js
@@ -85,7 +85,7 @@ exports.getStateResource = function * getState () {
 
   let signatureType = this.query.type || 'ed25519-sha512'
   let transfer = yield Transfer.findById(id)
-  let transferState = transfer ? transfer.state : null
+  let transferState = transfer ? transfer.state : 'nonexistent'
 
   let message = {
     id: uri.make('transfer', id),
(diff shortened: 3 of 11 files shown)


justmoon:
Please rename `identity` -> `connector` and rebase against master branch (that should fix CircleCI), otherwise LGTM.

sentientwaffle:
@emschwartz are you sure it should be added to https://github.com/interledger/five-bells-shared/blob/master/schemas/TransferState.json? The schemas are used to validate objects, and if we ever saw a Transfer with the `nonexistent` state, I would think we'd want to reject it.

emschwartz:
I guess the schemas kind of have two purposes -- in one sense they define
all of the valid possibilities for our objects and they're also used for
validation. Technically though, the transfer state receipt you returned is
now invalid according to this schema:
https://github.com/interledger/five-bells-shared/blob/master/schemas/TransferStateReceipt.json#L20

On Wed, Jan 6, 2016 at 4:45 PM, djg <notifications@github.com> wrote:

> @emschwartz <https://github.com/emschwartz> are you sure it should be
> added to
> https://github.com/interledger/five-bells-shared/blob/master/schemas/TransferState.json?
> The schemas are used to validate objects, and if we ever saw a Transfer
> with the nonexistent state, I would think we'd want to reject it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/interledger/five-bells-ledger/pull/84#issuecomment-169363219>
> .
>



-- 
Evan Schwartz | Software Architect | Ripple
[image: ripple.com] <http://ripple.com>

sentientwaffle:
Good point. Schema updates: https://github.com/interledger/five-bells-shared/pull/42

migrations/20150820161813-init.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
       primary: { type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true },
       name: { type: Sequelize.STRING, unique: true },
       balance: Sequelize.DECIMAL(10, 2),
-      identity: Sequelize.STRING(1024),
+      connector: Sequelize.STRING(1024),
```
emschwartz: You shouldn't change the migration, you should just add another one

src/controllers/accounts.js
```diff
@@ -14,6 +14,13 @@ exports.getCollection = function * find () {
   this.body = _.invoke(accounts, 'getDataExternal')
 }
 
+exports.getConnectors = function * () {
+  const accounts = yield Account.findAll({
+    where: { connector: { $ne: null } }
```
emschwartz: Is it possible for the connector value to be `undefined` or `''`?

src/controllers/transfers.js
```diff
@@ -65,12 +66,13 @@ exports.getResource = function * fetch () {
  * @apiVersion 1.0.0
  *
  * @apiDescription Use this to get a signed receipt containing only the id of
- *   transfer and its state.
+ *   transfer and its state. It functions even if the transfer doesn't exist yet.
```
emschwartz: Might be worth explicitly saying that if the transfer does not exist yet it will have the state set to `nonexistent`

emschwartz: Also, that case is only really useful for the hash pre-image signature scheme, right? @justmoon

src/controllers/accounts.js
sentientwaffle: It isn't possible for it to be `undefined`. It could be `''` if it was explicitly set to that by the admin, but I think that's on them.

emschwartz:
LGTM

src/controllers/transfers.js
justmoon: I think probably the `/state` endpoint should just return 404 if the transfer does not exist and you didn't pass the `pretend_state` query parameter.

Am I missing something? @emschwartz @sentientwaffle 

I definitely don't like the idea that the ledger will sign something other than the four canonical states of a two-phase commit. Theoretically, signing "nonexistent" could be used to prove that a transfer did *not* happen, so I guess it's useful...

sentientwaffle: Once we have a dedicated endpoint for getting the key and signature type that will be fine, but for now we need to be able to get the public_key and signature type before the transfer has been created.

clark800:
What was the reasoning for renaming to "connector"?

src/controllers/accounts.js
```diff
@@ -14,6 +14,13 @@ exports.getCollection = function * find () {
...
+  })
+  this.body = _.invoke(accounts, 'getDataConnector')
```
clark800: Should this be called `getConnectorData`?

justmoon:
@clark800 For pathfinding we need to be able to walk the graph, so the ledger should have a `/connectors` endpoint which lists all the connectors attached to a ledger. By setting the `connector` field on its account, a connector can signal to a ledger that it is a connector and what its URI is.

src/controllers/accounts.js
justmoon: We have previously talked about a naming scheme for these controller methods:

For a controller called `quibbles.js` and a model called `Quibble`:

`getResource` - Get a Quibble by primary key, usually GET /quibbles/:id
`getResourceBy[...]` - Get a Quibble by some other key, usually GET /quibbles/something/:foo
`getCollection` - Get a bunch of Quibbles, usually GET /quibbles
`get[...]Resource` - Get a resource related to Quibble by the Quibble's primary key, usually GET /quibbles/:id/something
`get[...]ResourceBy[...]` - Get a resource related to Quibble by some other index, usually GET /quibbles/something/:foo/something

So this could be `getConnectorCollection` perhaps.

We never really standardized on this scheme though and I don't care enough to worry about this. This name is only used in one place internally in the same project, so it's trivial to change anytime we want.

src/controllers/transfers.js
justmoon: Got it!

src/controllers/accounts.js
clark800: Ok, let's change it at the next convenient opportunity. This makes me wonder what a "data connector" is.

clark800:
Ok thanks

PR #85, opened 2016-01-05
Prevent duplicate subscriptions

### LINKED ISSUE #70: Prevent duplicate subscriptions
Multiple subscriptions with the same `event`, `subject` and `target` should be rejected with a 422 error code.

MatthewPhinney:
@justmoon been hacking on the repo to get up to speed ... I'd like to resolve #70 with this PR. Let me know if it looks ok.

### DIFF
--- a/migrations/20150820161813-init.js
+++ b/migrations/20150820161813-init.js
@@ -49,6 +49,8 @@ module.exports = {
       target: Sequelize.STRING(1024),
       created_at: Sequelize.DATE,
       updated_at: Sequelize.DATE
+    }).then(function () {
+      queryInterface.addIndex('Subscriptions', ['event', 'subject', 'target'])
     })
 
     queryInterface.createTable('Notifications', {
--- a/src/controllers/subscriptions.js
+++ b/src/controllers/subscriptions.js
@@ -7,6 +7,7 @@ const request = require('five-bells-shared/utils/request')
 const NotFoundError = require('five-bells-shared/errors/not-found-error')
 const Account = require('../models/account').Account
 const Subscription = require('../models/subscription').Subscription
+const UnprocessableEntityError = require('five-bells-shared/errors/unprocessable-entity-error')
 
 /**
  * Validate a subscription semantically.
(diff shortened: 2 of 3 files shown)


emschwartz:
LGTM

justmoon:
This query will quickly run into performance issues because it has to do a full table scan across the subscriptions table. You should create an index on those three columns.

One gotcha: Unfortunately there is an issue with Sequelize, MySQL and indices over large VARCHAR columns. MySQL has a size limit on indices, so you'll get an ER_TOO_LONG_KEY when you try to create that key. Normally you'd just specify that the index should only contain a prefix of the full VARCHAR columns, but there isn't any way in the Sequelize `addIndex` API to do that.

The solution I believe would be to add the index to the migration with a special case for MySQL.

Please try adding the index to the migration and let me know if you need any help.

MatthewPhinney:
@justmoon thanks for thinking about performance. I don't have much experience with mysql, but I installed it and ran the migration on a fresh mysql db without errors.  Here are the resulting indices in the Subscriptions table.
```
mysql> show index from Subscriptions;
+---------------+------------+------------------------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table         | Non_unique | Key_name                           | Seq_in_index | Column_name | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+---------------+------------+------------------------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| subscriptions |          0 | PRIMARY                            |            1 | id          | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| subscriptions |          1 | subscriptions_event_subject_target |            1 | event       | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
| subscriptions |          1 | subscriptions_event_subject_target |            2 | subject     | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
| subscriptions |          1 | subscriptions_event_subject_target |            3 | target      | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
+---------------+------------+------------------------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
```
Are there other caveats I should be aware of, or does this look OK?

justmoon:
> Are there other caveats I should be aware of, or does this look OK?

Nope, that LGTM

PR #87, opened 2016-01-06
[FEATURE] LEDGER_AUTH_{BASIC,HTTP_SIGNATURE}_ENABLED

### LINKED ISSUE #86: Ability to turn on/off auth schemes
It would be good to turn off HTTP Basic auth entirely via the configuration.

`LEDGER_AUTH_BASIC_ENABLED=0` - Turn off Basic auth
`LEDGER_AUTH_HTTP_SIGNATURE_ENABLED=0` - Turn off http-signature based auth

sentientwaffle:
Fixes #86.

### DIFF
--- a/src/services/config.js
+++ b/src/services/config.js
@@ -29,6 +29,8 @@ if (admin_pass) {
 }
 
 config.auth = {
-  basic_enabled: config.getEnv('AUTH_BASIC_ENABLED') !== '0',
-  http_signature_enabled: config.getEnv('AUTH_HTTP_SIGNATURE_ENABLED') !== '0'
+  basic_enabled: !isFalse(config.getEnv('AUTH_BASIC_ENABLED')),
+  http_signature_enabled: !isFalse(config.getEnv('AUTH_HTTP_SIGNATURE_ENABLED'))
 }
+
+function isFalse (val) { return val === '0' || val === 'false' }
--- a/src/services/config.js
+++ b/src/services/config.js
@@ -29,8 +29,13 @@ if (admin_pass) {
 }
 
 config.auth = {
-  basic_enabled: !isFalse(config.getEnv('AUTH_BASIC_ENABLED')),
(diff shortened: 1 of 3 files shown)


MatthewPhinney:
LGTM

src/services/config.js
```diff
@@ -27,3 +27,8 @@ if (admin_pass) {
     pass: admin_pass
   }
 }
+
+config.auth_enabled = {
```
clark800: `auth_enabled` kind of sounds like it should be a boolean

sentientwaffle: The values are. They are configured as strings because all env variables are strings.

clark800: Right, I just meant that in the code "config.auth_enabled" would look like a boolean, but I'm not sure if there is a better option. "config.auth.basic_enabled" is one other option. Not a big deal though, not trying to block the merge, just an idea.

justmoon: Agree with Chris - `config.auth.basic_enabled` matches our (not super-strict, but let's make an effort) config naming scheme better: `config.[module].[option]`

sentientwaffle:
Ok, renamed.

sentientwaffle:
Ok, if the env variable is undefined, it will default to true. If its `'0'`, it is false. Anything else will exit.

src/services/config.js
```diff
@@ -27,3 +27,15 @@ if (admin_pass) {
...
 }
+
+config.auth = {
+  basic_enabled: getBoolean('AUTH_BASIC_ENABLED'),
+  http_signature_enabled: getBoolean('AUTH_HTTP_SIGNATURE_ENABLED')
+}
+
+function getBoolean (env) {
+  const val = config.getEnv(env)
+  if (val === undefined) return true
```
clark800: Should probably allow "1" here also

sentientwaffle:
Ok, `"1"` -> `true`

clark800:
LGTM

PR #90, opened 2016-01-08
Allow admin to disable/enable accounts

### LINKED ISSUE #75: Add ability for admin to disable accounts
Since account deletion would rewrite history, the alternative is to disable accounts. Disabled accounts are not able to send or receive transfers. They do not count as valid users who can authenticate. Without authentication they are unable to see their own balance, or change any properties of their account. Disabled accounts should not be included in the planned `/connectors` endpoint.

Existing transfers involving disabled accounts in their debits or credits may not be prepared. Transfers that are already prepared should complete normally, even when they involve disabled accounts.

Only admins can disable or re-enable accounts.

MatthewPhinney:
resolves #75

My concerns with this PR are:
1. Had to add a new file to check if accounts are disabled. When creating a transfer, the only place where accounts appear to be validated is `accountBalances.js`. This file exposes a clean API for balances (which has nothing to do with accounts being disabled), so I created a new file (`disabledAccounts.js`) for the logic.
2. The JSON returned from PUT/GET account includes a `disabled` key *only* when the user is an admin. Ordinary users should not see the disabled flag, but admin users need visibility into which accounts are enabled/disabled. Having the structure of the response change based on which type of user makes the request is not ideal.

Finally, we need a corresponding PR to five-bells-shared to modify the JSON schema for accounts to allow the `disabled` flag. If the above concerns are addressed, and we decide to merge this PR, I'll make the PR to update the schema.


src/controllers/accounts.js
```diff
@@ -42,7 +42,7 @@ exports.getResource = function * fetch () {
   id = id.toLowerCase()
   log.debug('fetching account ID ' + id)
 
-  let can_modify = this.req.user.name === id || this.req.user.is_admin
+  let can_modify = (this.req.user.name === id && !this.req.user.disabled) || this.req.user.is_admin
```
MatthewPhinney: This should probably be named `can_view`.

justmoon:
Should add a test that shows that you can't send money *to* a disabled account.
Should remove the ability for a disabled account to pass either authentication method (basic or signature).

MatthewPhinney:
@justmoon can you clarify the execution flow for me? Does the account that authorized the transfer have to be responsible for making the request to execute that transfer? I'm concerned about the following case for a transfer Alice -> Bob:
1. Funds are put into escrow (Alice authorizes transfer to Bob).
2. Alice's account is disabled.
3. Alice cannot execute the transfer, since her credentials no longer work.

I agree that credentials for disabled accounts should not work. What is the workaround for this case?

emschwartz:
Once a transfer is authorized the account holder's credentials are no
longer needed and arguably the transfer should go through even if the
account is disabled. All that's needed is the execution condition
fulfillment and that will be provided by the connector. If the ledger
doesn't make good on the transfer once funds are escrowed the connector
will lose money.

justmoon:
Correct, a transfer that's already prepared can always execute. That's an important guarantee for two-phase commits. And you need to authenticate to submit a fulfillment, so we should be all good. I should have time to review the PR later today.

clark800:
> Ordinary users should not see the disabled flag

Why?

src/controllers/transfers.js
```diff
@@ -409,6 +410,8 @@ exports.putResource = function * create () {
       // version, but only allowing specific fields to change.
       transfer = updateTransferObject(originalTransfer, transfer)
     } else {
+      // If any accounts involved in a new transfer are disabled, throw error
+      yield hasDisabledAccounts(transaction, transfer)
```
clark800: The naming of this function indicates that it returns a boolean. Might want to call it something like `validateNoDisabledAccounts`.

src/lib/disabledAccounts.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+const _ = require('lodash')
+const Account = require('../models/account').Account
+const UnprocessableEntityError =
+require('five-bells-shared/errors/unprocessable-entity-error')
+
+function * hasDisabledAccounts (transaction, transfer) {
+  let accounts = _.groupBy(transfer.debits.concat(transfer.credits), function (creditOrDebit) {
```
clark800: Arrow function would look good here.

clark800: prefer const

clark800: map+uniq is semantically closer to what you are doing here than groupby

src/controllers/transfers.js
```diff
@@ -409,6 +410,8 @@ exports.putResource = function * create () {
       // version, but only allowing specific fields to change.
       transfer = updateTransferObject(originalTransfer, transfer)
     } else {
+      // If any accounts involved in a new transfer are disabled, throw error
```
clark800: I think this comment is redundant once the function is renamed

clark800:
> Ordinary users should not see the disabled flag

Oh, because they will get a 403 error if it's disabled anyways, so it's redundant right?

src/controllers/accounts.js
```diff
@@ -88,15 +95,18 @@ exports.putResource = function * putResource () {
...
-      yield Account.createExternal(self.body, { transaction })
+      yield Account.createExternal(account, { transaction })
     }
   })
 
   log.debug((existed ? 'updated' : 'created') + ' account name ' + name)
 
   this.body = this.body.getDataExternal()
+  if (!this.req.user.is_admin) {
+    delete this.body.is_disabled
```
justmoon: Don't really like the idea of having too many different versions of the schema. There is the user object, the public user object, the public user object without `is_disabled`...

My vote is not to delete the `is_disabled` property here. Makes it easier to write clients for the API and isn't hurting anybody.

src/controllers/accounts.js
```diff
@@ -54,6 +58,9 @@ exports.getResource = function * fetch () {
 
   this.body = can_examine ? account.getDataExternal() : account.getDataPublic()
   this.body.ledger = config.server.base_uri
+  if (!this.req.user || (this.req.user && !this.req.user.is_admin)) {
+    delete this.body.is_disabled
```
justmoon: Don't delete `is_disabled` - see other comment.

MatthewPhinney:
Thanks for the feedback. I made the suggested changes, with the exception of using arrow functions. The reason I didn't use them is that I didn't see them used elsewhere in the code, and I wanted to keep things consistent. I agree that arrow functions are generally cleaner, so I'm happy to switch to them, although doing so should probably be part of a larger refactor.

clark800:
I don't think arrow functions are the type of thing were going to have an explicit task for; it's probably best to just gradually change to arrow functions where appropriate. I think it's fine to do it here.

src/lib/disabledAccounts.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+const _ = require('lodash')
+const Account = require('../models/account').Account
+const UnprocessableEntityError =
+require('five-bells-shared/errors/unprocessable-entity-error')
+
+function * validateNoDisabledAccounts (transaction, transfer) {
+  const accounts = _.groupBy(transfer.debits.concat(transfer.credits), function (creditOrDebit) {
```
clark800: What about map+uniq?

MatthewPhinney: Again, this was just for consistency. See <a href="https://github.com/interledger/five-bells-ledger/blob/master/src/lib/accountBalances.js#L27">here</a>. But you're probably right that it makes sense to change the style as we go. I'll update the PR.

MatthewPhinney:
@clark800 updated and rebased.

clark800:
LGTM

PR #92, opened 2016-01-11
[FEATURE] part_of_payment -> additional_info

sentientwaffle:
Requires https://github.com/interledger/five-bells-shared/pull/44

### DIFF
--- a/src/models/transfer.js
+++ b/src/models/transfer.js
@@ -116,7 +116,7 @@ PersistentModelMixin(Transfer, sequelize, {
   ledger: Sequelize.STRING(1024),
   debits: JsonField(sequelize, 'Transfer', 'debits'),
   credits: JsonField(sequelize, 'Transfer', 'credits'),
-  part_of_payment: Sequelize.STRING(1024),
+  additional_info: JsonField(sequelize, 'Transfer', 'additional_info'),
   state: Sequelize.ENUM('proposed', 'prepared', 'executed', 'rejected'),
   execution_condition: JsonField(sequelize, 'Transfer', 'execution_condition'),
   execution_condition_fulfillment: JsonField(sequelize, 'Transfer', 'execution_condition_fulfillment'),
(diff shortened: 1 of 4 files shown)


src/models/transfer.js
```diff
@@ -116,7 +116,7 @@ PersistentModelMixin(Transfer, sequelize, {
   ledger: Sequelize.STRING(1024),
   debits: JsonField(sequelize, 'Transfer', 'debits'),
   credits: JsonField(sequelize, 'Transfer', 'credits'),
-  part_of_payment: Sequelize.STRING(1024),
+  additional_info: JsonField(sequelize, 'Transfer', 'additional_info'),
```
clark800: What do you think of the name "metadata"?

justmoon: Metadata in IT is "data about data" - hence metadata. For instance file metadata is the size of the data, creation date etc. This field is part of the data, so it's definitely not `metadata` - at a minimum metadata is something that you query separately from the data it describes.

I think the characteristic that we'd want to express is that the ledger is going to ignore whatever you put in this field. So it should be `custom_data` or `external_info` or something like that.

I think `additional_info` is actually not bad, because you should always design APIs from the perspective of the user and the user here is the developer who in her mind just wants to store some "additional fields" on the Transfer. So the name is something very intuitive to her. And of course consistency with Payment Services doesn't hurt.

clark800: In what sense is this `additional_info` part of the data? You could argue that file metadata is part of the file's data because they are stored together on the disk, but of course that's not the desired interpretation.

justmoon: > You could argue that file metadata is part of the file's data because they are stored together on the disk

If we're talking about actual filesystem metadata, that is not at all stored together with the data. In NTFS the metadata is stored in the MFT and in ext* it's stored in the [inode table](https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout#Inode_Table).

That's my point, if this data we're talking about was something that the ledger generated *about* the Transfer object, then I would call it metadata. As is, it's custom fields that you're adding into the object. 

All naming is ultimately a matter of taste. I would say unless somebody has a strong objection, I'm happy for us to merge the PR as-is.

LGTM.

PR #96, opened 2016-01-12
Return error message on invalid account name

MatthewPhinney:


MatthewPhinney:
We want to throw an error if the name in the body doesn't match the name in the URL param. Otherwise, if a user tries to create an account with a new name in the URL parameter, but accidentally gives the name of an existing account in the JSON body, they will get a 500 error coming from the DB update (due to a uniqueness constraint on account names).

clark800:
LGTM

emschwartz:
This should probably have had a test for that functionality

MatthewPhinney:
Yes, you may be right. I was on the fence about adding a unit test here. The reason I didn't was because we don't test for other 400 errors, and I didn't want to get into the mindset of adding a test for every possible edge case, since then the codebase quickly becomes unmanageable. In this case it may make sense though. A unit test could have uncovered this problem earlier, but we didn't test for it. Now that we know there is an issue, we may as well write the test.

clark800:
> I didn't want to get into the mindset of adding a test for every possible edge case

Actually I think we should be trying to add a test for every possible edge case :)

PR #103, opened 2016-01-13
Pr/160113101545

### LINKED ISSUE #99: Implement auth for subscriptions
We need to implement a permissions scheme so that only users who own an account can subscribe to events for that account.  For the code, see <a href="https://github.com/interledger/five-bells-ledger/blob/master/src/controllers/subscriptions.js#L25">here</a>.

@justmoon Should admin users be allowed to subscribe to events on any account?

MatthewPhinney:
Preliminary auth for subscription endpoints. The logic is as follows:
**GET**: a user may only view a subscription for an account they own, unless they are an admin, in which case they can view any subscription. They will receive a 404 if the subscription doesn't exist and a 403 if they are not authorized.
**PUT**: a user may only create a subscription in which the "owner" matches the user.
**What restrictions are placed on the `target` account(s)?**
**DELETE**: a user may only delete a subscription if they are the owner. An admin may delete any subscription. *Should we allow this?*

Once the above questions are addressed, this will resolve #99.

### DIFF
--- a/app.js
+++ b/app.js
@@ -60,7 +60,6 @@ router.put('/accounts/:id',
   models.Account.createBodyParser(),
   accounts.putResource)
 
-router.post('/subscriptions', models.Subscription.createBodyParser(), subscriptions.postResource)
 router.get('/subscriptions/:id',
   passport.authenticate(['basic', 'http-signature'], { session: false }),
   subscriptions.getResource)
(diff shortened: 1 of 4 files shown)


clark800:
What does the pull request name mean?

MatthewPhinney:
Looks like it's named after the branch: pr/`$(date +%y%m%d%H%M%S)`. The branch is created using some bash aliases Vahe and I worked with a while back. I can rename, if it's too cryptic.

clark800:
Descriptive names would be helpful

PR #104, opened 2016-01-14
Subscription auth

### LINKED ISSUE #99: Implement auth for subscriptions

MatthewPhinney:
Preliminary auth for subscription endpoints. The logic is as follows:
**GET**: a user may only view a subscription for an account they own, unless they are an admin, in which case they can view any subscription. They will receive a 404 if the subscription doesn't exist and a 403 if they are not authorized.
**PUT**: a user may only create a subscription in which the "owner" matches the user.
**What restrictions are placed on the `target` account(s)?**
**DELETE**: a user may only delete a subscription if they are the owner. An admin may delete any subscription. *Should we allow this?*

Once the above questions are addressed, this will resolve #99.

### DIFF
--- a/src/controllers/subscriptions.js
+++ b/src/controllers/subscriptions.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const uuid = require('uuid4')
 const db = require('../services/db')
 const log = require('../services/log')('subscriptions')
 const uri = require('../services/uriManager')
@@ -52,6 +51,7 @@ function * storeSubscription (subscription) {
  *
  * @apiUse NotFoundError
  * @apiUse InvalidUriParameterError
+ * @apiUse UnauthorizedError
  *
  * @returns {void}
  */
@@ -73,8 +73,8 @@ exports.getResource = function * fetch () {
 }
 
 /**
- * @api {post} /subscriptions Subscribe to an event
- * @apiName PostSubscription
+ * @api {put} /subscriptions Subscribe to an event
+ * @apiName PutSubscription
  * @apiGroup Subscription
  * @apiVersion 1.0.0
  *
@@ -87,26 +87,10 @@ exports.getResource = function * fetch () {
  *     }
  *
  * @apiUse InvalidBodyError
+ * @apiUse UnauthorizedError
  *
  * @returns {void}
  */
-exports.postResource = function * create () {
-  const subscription = this.body
-
-  // Generate a unique subscription ID outside of the transaction block
-  if (!subscription.id) {
-    subscription.id = uuid()
-  }
-  log.debug('preparing subscription ID ' + subscription.id)
-
-  // Validate and store subscription in database
-  yield * storeSubscription(subscription)
-
-  log.debug('subscription created')
-
-  this.body = subscription.getDataExternal()
-  this.status = 201
-}
 
 exports.putResource = function * update () {
   let id = this.params.id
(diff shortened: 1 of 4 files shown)


sentientwaffle:
This change requires a corresponding change in `five-bells-connector/lib/ledgers/five-bells-ledger.js`. Also, I can't think of a case where an admin would need to delete someone else's subscription.

justmoon:
> Also, I can't think of a case where an admin would need to delete someone else's subscription.

Admins should be able to do anything. The "admin" could be a script that is cleaning up abandoned subscriptions or a UI where you can view, edit and delete any subscriptions that exist on the ledger.

> a user may only view a subscription for an account they own

I only glanced at the code (sorry, time constrained at the moment), but I only saw checks on the `owner` property (who owns the subscription), not the `subject` property (which account is being listened to.) To prevent people from listening to other people's accounts, you need to check the `subject` property as well. Apologies if this already taken care of and I just missed it.

MatthewPhinney:
Thanks for the feedback. I updated the PUT endpoint so that admin users can have any account in the `subject` field. Ordinary users can only subscribe to their own accounts, so they may only have their account as the `subject` (`subject` and `owner` must match for normal users). The `owner` must always match the user in the HTTP request, regardless of whether the user is admin or not.

The GET/DELETE endpoints have not changed. Admins may GET or DELETE any subscription. All other users may GET/DELETE only subscriptions they own.

Please let me know if this looks OK.

@sentientwaffle what needs to be changed in the connector? Is it just the auth for the `subscribe`/`unsubscribe` methods? I'm not familiar with that code.

sentientwaffle:
@MatthewPhinney ya, right now `subscribe`/`unsubscribe` don't send any credentials.

MatthewPhinney:
Incidentally, I noticed that `subject` is a string, while `owner` is an IRI in the schema <a href="https://github.com/interledger/five-bells-shared/blob/master/schemas/Subscription.json#L13">here</a>. Do we want `subject` to ref IRI too? Presumably they are both accounts.

justmoon:
@MatthewPhinney Yeah `subject` and `target` should be URIs.

MatthewPhinney:
Fixed conflicts and rebased. I'd like to give this another look before it's merged.

src/controllers/subscriptions.js
```diff
@@ -80,17 +61,20 @@ exports.getResource = function * fetch () {
...
   log.debug('fetching subscription ID ' + id)
 
+  const requestOwner = uri.make('account', this.req.user.name)
   const subscription = yield Subscription.findById(id)
-  if (subscription) {
-    this.body = subscription.getDataExternal()
-  } else {
+  if (!subscription) {
     throw new NotFoundError('Unknown subscription ID')
+  } else if (!(requestOwner === subscription.owner || this.req.user.is_admin)) {
```
clark800: This pattern repeats a few times, maybe we should make a function `isAuthorized(this.req.user, subscription.owner/subscription.subject)`

MatthewPhinney:
@clark800 I think this is good to go. I added a few helper methods, as you suggested. They don't belong in the controller, but we can refactor them as part of the larger refactor task.

clark800:
LGTM

PR #106, opened 2016-01-14
[FEATURE] Allow account+ledger lookup without auth

### LINKED ISSUE #105: Enable ledger to be looked up from account identifier
Right now when doing pathfinding we need to pass `source_account`, `source_ledger` and `source_username`, even though `source_account` combines the two other pieces of information. We're supposed to treat URIs as opaque so we can't assume that we can parse the username or ledger from the `source_account`.

My proposal would be to allow anyone (without authentication) to `GET` the account URI and return an object that contains the `ledger` and `account`.

One thing that would enable would be that your "account" that you share with people could actually be a service not tied to your ledger that would return your up to date payment info. If the standard behavior was for people to `GET` that endpoint first, you'd be able to update what that points to whenever you want to switch where you hold your money. This would be very useful for cases when you want to embed payment information in some data that can't be easily updated when you move accounts.

The first potential problem with this is that that endpoint would become a target for hackers because if they could change where that points to they could redirect where your money goes.

The second potential problem is that you might not want anyone to be able to find out if an account exists. That could be addressed by adding functionality to the ledgers to enable account holders to create aliases for their account that they could give to specific parties.

sentientwaffle:
Fixes #105 

Depends on https://github.com/interledger/five-bells-shared/pull/45

### DIFF
--- a/test/accountSpec.js
+++ b/test/accountSpec.js
@@ -28,6 +28,7 @@ describe('Accounts', function () {
     this.adminAccount = this.exampleAccounts.admin
     this.holdAccount = this.exampleAccounts.hold
     this.existingAccount = this.exampleAccounts.alice
+    this.existingAccount2 = this.exampleAccounts.candice
     this.traderAccount = this.exampleAccounts.trader
 
     // Reset database
@@ -38,6 +39,7 @@ describe('Accounts', function () {
       this.adminAccount,
       this.holdAccount,
       this.existingAccount,
+      this.existingAccount2,
       this.traderAccount
     ])
   })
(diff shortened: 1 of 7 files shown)


test/accountSpec.js
```diff
@@ -106,27 +110,38 @@ describe('Accounts', function () {
         .end()
     })
 
-    it('should return 401 when not authenticated', function *() {
+    it('should return 200/404 when not authenticated, with partial data', function *() {
```
MatthewPhinney: Maybe this could be two tests? One for the existing account and one for the nonexistent one?

sentientwaffle:
Ok, tests are split.

clark800:
LGTM

MatthewPhinney:
LGTM

PR #113, opened 2016-01-19
[TASK] Refactor app.js

### LINKED ISSUE #93: SQLite transaction compatibility code should be moved to five-bells-shared/lib/db.js
This code:

https://github.com/interledger/five-bells-ledger/blob/47717ff9d2292cc7b03fe9de5d5d43f31c943463/app.js#L85-L99

Should be moved to `five-bells-shared/lib/db.js` into a static generator method called `init`.

### LINKED ISSUE #94: Automatic account creation should be moved into a class
This code:

https://github.com/interledger/five-bells-ledger/blob/47717ff9d2292cc7b03fe9de5d5d43f31c943463/app.js#L101-L121

Should be moved to a class.

### LINKED ISSUE #95: App should be a class
This code:

https://github.com/interledger/five-bells-ledger/blob/47717ff9d2292cc7b03fe9de5d5d43f31c943463/app.js#L30-L128

Should live in a class called `App`, so that I can easily include the ledger as a module and still start it.

For an example, see the notary:
https://github.com/interledger/five-bells-notary/blob/db831cfabf49ef82947846ba87de019ff48c5c87/app.js
https://github.com/interledger/five-bells-notary/blob/ec60a2cff187f3bcb906fa3fd52826d65557108c/src/lib/app.js

sentientwaffle:
Requires https://github.com/interledger/five-bells-shared/pull/47

Fixes #93
Fixes #94
Fixes #95

/cc @justmoon


sentientwaffle:
Ok; added the new-and-improved interface.

src/lib/fixtures.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+const models = require('../models')
+
+function Fixtures (db, config) {
```
clark800: The definitions of "fixtures" that I can find all refer to testing. It's probably best to use another name to avoid confusion.

app.js
```diff
@@ -1,136 +1,7 @@
 /* @flow */
```
clark800: This looks like a source file. I think we should move it to /src.

src/lib/app.js
```diff
@@ -0,0 +1,111 @@
...
+const health = require('../controllers/health')
+const transfers = require('../controllers/transfers')
+const accounts = require('../controllers/accounts')
+const subscriptions = require('../controllers/subscriptions')
+const models = require('../models')
+
+// Configure passport
+require('../services/auth')
+
+function App (modules) {
```
clark800: Can this be an ES6 class?

src/lib/app.js
```diff
@@ -0,0 +1,111 @@
...
+  yield this.fixtures.setup()
+
+  this.app.listen(this.config.server.port)
+  this.log.info('ledger listening on ' +
+    this.config.server.bind_ip + ':' +
+    this.config.server.port)
+  this.log.info('public at ' + this.config.server.base_uri)
+}
+
+App.prototype.makeRouter = function () {
```
clark800: Should this be private?

src/lib/fixtures.js
clark800: I think this is a good example of when OOP causes design problems. The function `setupHoldAccount` has access to the `config` object even though it doesn't need it, which means we have introduced unnecessary coupling. Also `db` is never used in this class.

I think these should just be two regular functions without any class. Then the linter should warn you that you have unused arguments.

src/lib/app.js
```diff
@@ -0,0 +1,111 @@
...
+
+function App (modules) {
+  this.log = modules.log('app')
+  this.config = modules.config
+  this.db = modules.db
+  this.timerWorker = modules.timerWorker
+  this.notificationWorker = modules.notificationWorker
+  this.fixtures = modules.fixtures
+
+  const app = this.app = koa()
```
clark800: The `app.app` looks weird. Maybe we could do:

```
const Koa = require('koa');
...
this.koa = Koa();
```

src/lib/fixtures.js
```diff
@@ -0,0 +1,39 @@
...
+Fixtures.prototype.setupHoldAccount = function * () {
+  const holdAccount = yield models.Account.findByName('hold')
+  if (!holdAccount) {
+    yield models.Account.create({name: 'hold', balance: '0'})
+  }
+}
+
+Fixtures.prototype.setupAdminAccount = function * () {
+  if (!this.config.default_admin) return
+  let adminParams = this.config.default_admin
```
clark800: prefer const here and next line

sentientwaffle:
Ok, its ready for another look

clark800:
LGTM

Issue #112, opened 2016-01-15
Three-layer Architecture


clark800:
Split code into three layers:
1. Service layer: HTTP server and HTTP request/response handling
2. Domain layer: Business logic
3. Persistence layer: Database access
See: https://en.wikipedia.org/wiki/Multilayered_architecture

Currently there is not a clean segregation of responsibilities; `transfers.putResource` does all three in a single function. The directory structure can be refactored to:

1. /src/server - HTTP router, route handlers, HTTP parameter extraction, HTTP response formatting, converting return values and exceptions to HTTP status codes, etc
2. /src/domain - Business logic
3. /src/data - Database access

Layer 1 can depend on layer 2, layer 2 can depend on layer 3, but no other dependencies should be allowed.

MatthewPhinney:
This will be a worthwhile exercise, particularly as the code base grows.  I think the biggest issue to consider here is if we want to use the existing model class in five-bells-shared for the persistence layer.

justmoon:
Totally agree that the ledger controllers should not contain the business logic.  Express (and Koa) are MVC frameworks, so they really work best in conjunction with the MVC pattern. In MVC, business logic [should be handled in the model](http://stackoverflow.com/questions/18563229/mvc-where-to-put-business-logic).

In general I think there are a few common folder names for Express applications that we should adhere to, such as `models/`, `controllers/`, `views/` (if any), `middlewares/` (if any).

In terms of routers, I like the notary's setup. There's a Router class, which includes the controllers and has them register their routes. Finally, an "attach" method attaches the Router to the App.

https://github.com/interledger/five-bells-notary/blob/master/src/lib/router.js
https://github.com/interledger/five-bells-notary/blob/master/src/controllers/cases.js

Here are some resources on what others are doing:

https://www.terlici.com/2014/08/25/best-practices-express-structure.html
https://gist.github.com/lancejpollard/1398757
http://stackoverflow.com/questions/5178334/folder-structure-for-a-node-js-project

Note how MVC is pretty much universally used - it has worked well for me in the past and most engineers are familiar with it.

clark800:
MVC is an architecture that is designed to separate the user interface from the business logic. Since this project doesn’t have a user interface (view), it is a bit of a square peg in a round hole situation.

If we want to use MVC for this project, we can really only use the “model” and “controller” parts because we don’t have any views. But consider some of the definitions of these terms:

[...]

PR #116, opened 2016-01-21
Refactor accounts controller

clark800:
I'm only refactoring "accounts" in this pull request so that everyone has a chance to give feedback before I proceed with the rest of the controllers.


src/controllers/accounts.js
```diff
@@ -1,26 +1,15 @@
...
-const UnauthorizedError =
-require('five-bells-shared/errors/unauthorized-error')
-const Account = require('../models/account').Account
+const model = require('../models/accounts')
 
-exports.getCollection = function * find () {
-  const accounts = yield Account.findAll()
-  this.body = _.invoke(accounts, 'getDataExternal')
+function * getCollection () {
+  this.body = yield model.getAccounts()
```
MatthewPhinney: What are `getCollection` and `getConnectors` used for?

src/controllers/accounts.js
```diff
@@ -75,32 +48,20 @@ exports.getResource = function * fetch () {
...
-
-  log.debug((existed ? 'updated' : 'created') + ' account name ' + name)
-
-  this.body = this.body.getDataExternal()
-  this.status = existed ? 200 : 201
+module.exports = {
+  getCollection,
+  getConnectors,
+  getResource,
+  putResource
```
MatthewPhinney: Why export an object instead of  `exports.putResource = function`? Personally, I prefer exporting an object as well, but I am curious if you have specific reasons for doing so.

clark800: There is less redundancy, less clutter near the actual code, and I like the symmetry and organization of all the imports at the top and all the exports at the bottom.

src/controllers/accounts.js
clark800: Listing all accounts and listing all connectors (accounts for FX connectors)

MatthewPhinney: I don't see the `/connectors` endpoint in the doc. Is this documented somewhere?

clark800: Looks like there's no apidoc, so probably not.

MatthewPhinney:
As per the discussion in #112, we'll want to refactor the domain/persistence layer as well, but as far as the controller/model separation is concerned, I think this refactor does the job.

MatthewPhinney:
LGTM

lumberj:
LGTM

PR #119, opened 2016-01-25
[FIX] allow admin to subscribe to any account

sentientwaffle:
This is necessary to fix the visualization. /cc @MatthewPhinney


clark800:
LGTM

Issue #83, opened 2016-01-01
Notification IDs are leaking information


justmoon:
After #80, notification IDs will be public. The notification ID is currently a serial integer. This means it's leaking information about the number of notifications that are happening.

We should change that to some kind of unique ID. Could be a UUID or a shorter time-based ID like [Snowflake](https://www.npmjs.com/package/flake-idgen).

rec:
What about a hash of the notification contents and timestamp?

justmoon:
That would work as well.

Ideally, we want something that's unique and as small as possible. For the timestamp you need at least 32 bits, better a little more. Snowflake uses a 42-bit timestamp, a 10 bit node id and a 12-bit counter. This ensures uniqueness and only costs 64 bits total. In a distributed system it's very hard/expensive to create a counter, but by adding the 10-bit node id, you basically have a separate counter for each running worker. And by putting the timestamp first, we can use the ID both for identification as well as sorting (i.e. give me the most recent notification.) 

Using a hash is pretty much equivalent to using random bits. So for instance we could use a 32-bit timestamp and a 32-bit hash. Collisions would be pretty unlikely, but not to the "astronomically unlikely" degree that for instance a UUIDv4 (with 126 bits of randomness) gives you.

Twitter's solution (Snowflake) is minimal, elegant, available as a preexisting module and makes collisions impossible (not just unlikely.) So it seems like a good option.

rec:
A 10-bit node ID means a maximum of 1024 nodes, which might not be a desirable limit to have hardcoded...?

MatthewPhinney:
Where is the notification ID a serial integer? Is this a key in some DB? #80 calls for using an identifier, but it looks like the identifier is generated by the ledger, in which case we could use any unique ID (read: Snowflake, not serial int). Can we combine this issue and #80 into a single issue?

MatthewPhinney:
Never mind, my local copy was out of date ... I didn't have the `notifications` table.

MatthewPhinney:
https://github.com/MatthewPhinney/five-bells-ledger/commit/da97062488478a9b4925feac0a2fb9b35ce91e8b

Issue #80, opened 2015-12-30
Notifications should have their own ID


justmoon:
Current:

``` js
{
  "id": "http://eur-ledger.example/subscriptions/52a42d6f-8d9c-4c05-b31c-cccc8bbdb30d",
  "event": "transfer.update",
  "resource": {
    // ...
  }
}
```

Proposed:

``` js
{
  "id": "http://eur-ledger.example/subscriptions/52a42d6f-8d9c-4c05-b31c-cccc8bbdb30d/notifications/140",
  "subscription": "http://eur-ledger.example/subscriptions/52a42d6f-8d9c-4c05-b31c-cccc8bbdb30d",
  "event": "transfer.update",
  "resource": {
    // ...
  }
}
```

Edit: Changed notification URI to start with subscription URI.

MatthewPhinney:
https://github.com/MatthewPhinney/five-bells-ledger/commit/271f193de6e094d9f55f294fa50eba0ed2239671

PR #120, opened 2016-01-25
Add ID for notifications

MatthewPhinney:


MatthewPhinney:
Resolve #80. Resolve #83.

test/putTransferSpec.js
```diff
@@ -974,9 +974,11 @@ describe('PUT /transfers/:id', function () {
 
     const notification = nock('http://subscriber.example')
       .post('/notifications', (body) => {
+        const idParts = body.id.split('/')
+        const notificationId = body.id.split('/')[idParts.length - 1]
```
clark800: Looks like you can use the `idParts` here to avoid splitting a second time

MatthewPhinney: Oops, meant to fix that earlier ....

clark800:
Can you add a test case?

clark800:
Is it possible to GET an individual notification by id?

MatthewPhinney:
As far as I know, we do not support that. @justmoon do we intend to support querying notifications by ID?

justmoon:
> @justmoon do we intend to support querying notifications by ID?

Yes, regarding the authentication of certificates, I would propose that we give people two options:

1. Validate the TLS client certificate when the notification is submitted. - Widely supported, standards-based solution.

2. GET the notification from the origin server. - Some developer may have difficulty with method 1. Perhaps their TLS implementation doesn't support validating client certs very well. So as an alternative, they can GET the notification from the origin server simply by doing `fetch(notification.id).then(body => if (deepEquals(notification, body)) { ... })`.

Pattern 2 is less efficient, but extremely simple. One implementation note: User should GET the notification (and process it) **before** responding positively to the PUT request. After the receiving server has acknowledged the PUT, the sending server (in this case the ledger) should be free to delete their record of the notification if they want.

MatthewPhinney:
@clark800 what kind of tests did you have in mind? The PR only changes the format of the notification, and I updated the existing test to expect this new format.

clark800:
I was looking for a test of GET, then I noticed that it wasn't implemented

MatthewPhinney:
@clark800 can you please review this?

clark800:
What about the GET endpoint?

MatthewPhinney:
I can add that, but I don't see any information about it in the design docs. Maybe we could track it with a separate JIRA issue that has more design details?

clark800:
It's just the REST model, if you have a URI it should actually "exist" in the sense of having some kind of behavior at that URI.

emschwartz:
+1 the whole point of adding IDs is to allow clients to `GET` them from the ledger.

See Stefan's comment from above:
> GET the notification from the origin server. - Some developer may have difficulty with method 1. Perhaps their TLS implementation doesn't support validating client certs very well. So as an alternative, they can GET the notification from the origin server simply by doing fetch(notification.id).then(body => if (deepEquals(notification, body)) { ... }).

MatthewPhinney:
That makes sense, I was just thinking we could have a design discussion elsewhere. But we may as well do it here. Should the response to GET `/subscriptions/:subscription_id/notifications/:notification_id` be the same as the payload of the POST notification that the ledger performs with queued notifications?
```
{
  "id": "http://eur-ledger.example/subscriptions/52a42d6f-8d9c-4c05-b31c-cccc8bbdb30d/notifications/fccacc64-8b8d-4dd6-91f5-8ffc7674e8be",
  "subscription": "http://eur-ledger.example/subscriptions/52a42d6f-8d9c-4c05-b31c-cccc8bbdb30d",
  "event": "transfer.update",
  "resource": {
    // ...
  }
}
```
I've implemented most of this but want to make sure we're all on the same page before flushing out the code and adding proper tests.

clark800:
> Should the response to GET /subscriptions/:subscription_id/notifications/:notification_id be the same as the payload of the POST notification that the ledger performs with queued notifications?

Is there a reason that it should be different?

src/models/notifications.js
```diff
@@ -0,0 +1,43 @@
...
+function * getNotification (subscriptionId, notificationId, requestingUser) {
+  log.debug('fetching notification ID ' + notificationId)
+  const notification = yield Notification.findOne({
+    where: {
+      id: notificationId
+    },
+    include: [{ all: true }]})
+  if (!notification) {
+    throw new NotFoundError('Unknown notification ID')
+  } else if (notification.Subscription.id !== subscriptionId) {
```
MatthewPhinney: This will throw if `notification.Subscription` is `undefined`, which should never happen, but is possible if the database becomes corrupted. I can add a check for this case, if we think it's worthwhile.

src/models/notifications.js
```diff
@@ -0,0 +1,43 @@
+'use strict'
+
+const log = require('../services/log')('subscriptions')
+const uri = require('../services/uriManager')
+const NotFoundError = require('five-bells-shared/errors/not-found-error')
+const UnauthorizedError = require('five-bells-shared/errors/unauthorized-error')
+const Notification = require('./db/notification').Notification
+const Transfer = require('./db/transfer').Transfer
+
+function isOwnerOrAdmin (requestingUser, subscription) {
```
MatthewPhinney: This method is taken from the `subscriptions` model. It may make sense to put it in a separate file of library methods that both the `notifications` model and the `subscriptions` model can access, to avoid code duplication.

MatthewPhinney:
The GET notifications endpoint has been implemented.

test/data/notificationExample.json
```diff
@@ -0,0 +1,5 @@
+{
```
clark800: I think this is a database entry right? At first I interpreted this as a notification response. Maybe rename to something like "notificationDatabaseEntry.json"?

src/models/notifications.js
clark800: Seems worthwhile

src/models/notifications.js
clark800: I'm all for reducing duplication

MatthewPhinney:
I moved the subscription util methods into a separate file, so that we can use them in both the `notifications` model and the `subscriptions` model. I also added a check to ensure the notification has a corresponding subscription in the DB. I did not add a test for this specific case, since doing so requires adding an example notification which has a subscription ID for a non-existent subscription. Doing this causes a FK constraint error.

src/models/notifications.js
```diff
@@ -0,0 +1,41 @@
...
+function * getNotification (subscriptionId, notificationId, requestingUser) {
+  log.debug('fetching notification ID ' + notificationId)
+  const notification = yield Notification.findOne({
+    where: {
+      id: notificationId
+    },
+    include: [{ all: true }]})
+  if (!notification) {
+    throw new NotFoundError('Unknown notification ID')
+  } else if (notification.Subscription === undefined) {
```
clark800: I'm not sure how sequelize works here, would it be `undefined` or `null`?

MatthewPhinney: Good point, I checked and it's `null`. I updated the code to reflect this.

clark800:
LGTM

PR #122, opened 2016-01-25
Refactor controllers - move business logic to models

clark800:


clark800:
Note: removed `jsonld.setContext` call; it looked like it wasn't being used.

src/controllers/transfers.js
```diff
@@ -359,126 +129,27 @@ function * processStateTransitions (tr, transfer) {
...
-    if (!transfer.isFinalized()) {
-      yield transferExpiryMonitor.watch(transfer)
-    }
-  })
-
-  log.debug('changes written to database')
-
-  this.body = transfer.getDataExternal()
-  this.status = originalTransfer ? 200 : 201
+  transfer.id = id.toLowerCase()
```
MatthewPhinney: You only need to set this if the above `if` block does not execute. Maybe wrap this in an `else` statement, for consistency with `controllers/subscriptions.js`.

src/models/transfers.js
```diff
@@ -0,0 +1,354 @@
...
+    transferStateReceipt.digest = sha256(stringifyJSON(realPreImage))
+    transferStateReceipt.message = makePreImage(message.id, transferState)
+    if (conditionState) {
+      const conditionPreImage = makePreImage(message.id, conditionState)
+      transferStateReceipt.condition_state = conditionState
+      transferStateReceipt.condition_digest = sha256(stringifyJSON(conditionPreImage))
+    }
+  } else {
+    // TODO: move this validation to controller or remove "URI"
+    throw new InvalidUriParameterError('type is not valid')
```
MatthewPhinney: I think this can go in the controller. Provided we only support a few algorithms, we can just validate that the URI param is one of them. If the list grows, we could check against some kind of system dictionary.

src/models/transfers.js
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  return transferStateReceipt
+}
+
```
MatthewPhinney: Do you think it makes sense to have a separate file for crypto methods?

src/controllers/transfers.js
clark800: Well you still have to set it in case `transfer.id` is uppercase. I changed the subscriptions code to match this.

src/models/transfers.js
clark800: Probably, but I think that's a separate refactoring

src/models/transfers.js
```diff
@@ -0,0 +1,354 @@
...
+  transfer.executed_at = updatedTransferData.executed_at
+  transfer.rejected_at = updatedTransferData.rejected_at
+
+  // Ignore undefined properties
+  const transferData = _.omit(transfer.getData(), _.isUndefined)
+
+  // Clients can add authorizations
+  // The validity of these authorizations will be checked
+  // in the validateAuthorizations function
+  _.forEach(updatedTransferData.debits, function (funds, i) {
```
MatthewPhinney: Could use an arrow function here.

src/models/transfers.js
```diff
@@ -0,0 +1,354 @@
...
+      // a security vulnerability?
+      throw new UnauthorizedError('Unknown or invalid account / password')
+    }
+  }
+
+  // Go through each of the debits in the transfer
+  // Throw an error if authorized is set to true on any debit
+  // that was not set to true on the transferFromDb and is not
+  // owned by the authorizedAccount
+  _.forEach(transfer.debits, function (debit, debitIndex) {
```
MatthewPhinney: Arrow function?

src/models/transfers.js
```diff
@@ -0,0 +1,354 @@
...
+    if (debit.account !== authorizedAccount &&
+      (!previousDebits || !previousDebits[debitIndex].authorized)) {
+      throw new UnauthorizedError('Invalid attempt to authorize debit')
+    }
+  })
+
+// TODO: add credit authorization
+}
+
+function * processStateTransitions (tr, transfer) {
```
MatthewPhinney: I guess `tr` is a db transaction, but I think it would be a lot clearer if this was spelled out (rename `tr` to `transaction`).

src/models/transfers.js
clark800: This code is just moved from controllers, low-level refactoring of the copied code is out of scope for this pull request.

src/models/transfers.js
clark800: Updated

MatthewPhinney:
LGTM

PR #123, opened 2016-01-26
Extract database code from accounts model

clark800:
Moves lower-level database access from the models to a database abstraction layer so that the business logic in the model is easier to read.

### DIFF
--- a/src/models/db/accounts.js
+++ b/src/models/db/accounts.js
@@ -1,0 +1,42 @@
+'use strict'
+
+const Account = require('./account').Account
+const db = require('../../services/db')
+
+function * getAccounts () {
+  return (yield Account.findAll())
+}
+
+function * getConnectorAccounts () {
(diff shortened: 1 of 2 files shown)


src/models/db/accounts.js
```diff
@@ -0,0 +1,42 @@
...
+  yield db.transaction(function * (transaction) {
+    existed = yield Account.findByName(account.name, { transaction })
+    if (existed) {
+      existed.setDataExternal(account)
+      yield existed.save({ transaction })
+    } else {
+      yield Account.createExternal(account, { transaction })
+    }
+  })
+  return existed
```
MatthewPhinney: I think it makes sense to return a boolean here for consistency with sequelize's `upsert` method.

clark800: Ok, updated

MatthewPhinney:
LGTM

PR #124, opened 2016-01-26
Store hashed passwords in database

clark800:
Depends on https://github.com/interledger/five-bells-shared/pull/51

### DIFF
--- a/src/lib/seed-db.js
+++ b/src/lib/seed-db.js
@@ -1,4 +1,5 @@
 'use strict'
+const bcrypt = require('bcrypt')
 const models = require('../models/db')
 
 module.exports = function * (config) {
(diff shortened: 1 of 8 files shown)


clark800:
This is a breaking change because account passwords are set by PUT'ing an account, which now expects a `password_hash` field instead of a `password` field.

src/lib/seed-db.js
```diff
@@ -15,18 +16,31 @@ function * setupHoldAccount () {
   }
 }
 
+function * hashPassword (password) {
+  return yield new Promise((resolve, reject) => {
```
MatthewPhinney: What's the advantage of returning a promise here?

MatthewPhinney:
Don't you need to also update the db migrations to rename `password` -> `password_hash`?

src/lib/seed-db.js
MatthewPhinney: Never mind, I read the docs ...

clark800:
> Don't you need to also update the db migrations to rename password -> password_hash?

Yeah, I should add a migration, but it isn't just a renaming, so it won't be reversible.

MatthewPhinney:
LGTM

src/lib/seed-db.js
```diff
@@ -15,18 +16,31 @@ function * setupHoldAccount () {
...
+    bcrypt.genSalt(10, (error, salt) => {
```
lumberj: I had to lookup the meaning of `10` here and it's the number of rounds (defaults to 10). This is a case where I'd argue for an explaining variable (`const rounds = 10`) to pass into the function. But LGTM otherwise

clark800: Ok, updated

sentientwaffle:
Whats the advantage of forcing clients to PUT a password_hash instead of just hashing it on arrival? it forces all of the clients to depend on bcrypt

clark800:
There are two reasons I chose to do it this way:
1. It's more secure to hash the password as early as possible after the user inputs it.
2. It's more consistent with the concept of a "PUT" if the data that you "put" is the data that gets stored; transforming it in between adds another layer of complexity.

PR #125, opened 2016-01-26
Initial implementation for client certificate authentication

lumberj:
Depends on:

~~https://github.com/interledger/five-bells-shared/pull/50~~
~~https://github.com/interledger/five-bells-shared/pull/52~~

Working example:

Server
```diff
   _makeRouter () {
     const router = new Router()
-    router.get('/health', health.getResource)
+
+    router.get('/health',
+        passport.authenticate(['client-cert']),
+        health.getResource)

```

Client
```js
var fs = require('fs');
var https = require('https');

var options = {
  hostname: 'localhost',
  port: 3000,
  path: '/health',
  method: 'GET',
  key: fs.readFileSync('client1-key.pem'),
  cert: fs.readFileSync('client1-crt.pem'),
  ca: fs.readFileSync('ca-crt.pem') // Trust the test servers self-signed cert
};

var req = https.request(options, function(res) {
  res.on('data', function(data) {
    process.stdout.write(data);
  });
});
req.end();
```

### DIFF
--- a/migrations/20160127110439-account-fingerprint.js
+++ b/migrations/20160127110439-account-fingerprint.js
@@ -1,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    queryInterface.addColumn('Accounts', 'fingerprint', Sequelize.STRING)
+  },
+
+  down: function (queryInterface, Sequelize) {
+    queryInterface.removeColumn('Accounts', 'fingerprint')
+  }
+};
--- a/src/models/db/account.js
+++ b/src/models/db/account.js
@@ -70,6 +70,13 @@ class Account extends Model {
     })
   }
 
+  static findByFingerprint (fingerprint, options) {
+    return Account.findOne({
+      where: {fingerprint: fingerprint},
+      transaction: options && options.transaction
+    })
+  }
+
   createEntry (values, options) {
     values.account = this.primary
     values.balance = this.balance
@@ -109,7 +116,13 @@ PersistentModelMixin(Account, sequelize, {
   password_hash: Sequelize.STRING,
   public_key: Sequelize.TEXT,
   is_admin: Sequelize.BOOLEAN,
-  is_disabled: Sequelize.BOOLEAN
-})
+  is_disabled: Sequelize.BOOLEAN,
+  fingerprint: Sequelize.STRING
+},
+  {
+    indexes: [
+      {fields: ['fingerprint']}
+    ]
+  })
 
 exports.Account = Account
--- a/src/services/auth.js
+++ b/src/services/auth.js
@@ -3,6 +3,7 @@
 const bcrypt = require('bcrypt')
 const passport = require('koa-passport')
 const BasicStrategy = require('passport-http').BasicStrategy
+const ClientCertStrategy = require('five-bells-shared/middlewares/passport-client-cert')
 const HTTPSignatureStrategy = require('passport-http-signature')
 const AnonymousStrategy = require('passport-anonymous').Strategy
 const Account = require('../models/db/account').Account
@@ -53,6 +54,21 @@ passport.use(new HTTPSignatureStrategy(
         done(null, userObj, userObj.public_key)
       })
   }))
+
+passport.use(new ClientCertStrategy((fingerprint, info, done) => {
+  if (!config.getIn(['auth', 'client_certificates_enabled'])) {
+    return done(new UnauthorizedError('Unsupported authentication method'))
+  }
+
+  Account.findByFingerprint(fingerprint)
+    .then(function (userObj) {
+      if (!userObj || userObj.is_disabled || !userObj.fingerprint ||
+          !fingerprint || userObj.fingerprint !== fingerprint) {
+        return done(new UnauthorizedError('Unknown or invalid account'))
+      }
+      done(null, userObj)
+    })
+}))
 
 // Allow unauthenticated requests (transfers will just
 // be in the proposed state)
--- a/src/services/config.js
+++ b/src/services/config.js
@@ -21,8 +21,9 @@ if (admin_pass) {
 }
 
 localConfig.auth = {
-  basic_enabled: Config.castBool(Config.getEnv('AUTH_BASIC_ENABLED'), true),
-  http_signature_enabled: Config.castBool(Config.getEnv('AUTH_HTTP_SIGNATURE_ENABLED'), true)
+  basic_enabled: Config.castBool(Config.getEnv(envPrefix, 'AUTH_BASIC_ENABLED'), true),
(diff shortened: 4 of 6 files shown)


lumberj:
TODO:

- [ ] Configuration for passing in certificate, key, and ca
- [ ] Enable client-certificate auth for authenticated endpoints
- [ ] Documentation

src/services/auth.js
```diff
@@ -48,6 +49,20 @@ passport.use(new HTTPSignatureStrategy(
...
   }))
 
+passport.use(new ClientCertStrategy((userName, info, done) => {
+  if (!config.auth.client_certificates_enabled) {
+    return done(new UnauthorizedError('Unsupported authentication method'))
+  }
+
+  Account.findByName(userName)
+    .then(function (userObj) {
+      if (!userObj || userObj.is_disabled || userObj.fingerprint !== info.fingerprint) {
```
MatthewPhinney: Do you need to add the `fingerprint` to the account object?

lumberj: There is still some discussion over whether the `userName` will be the fingerprint. For now, we can assume we're either using the fingerprint itself as the username or, the Common Name as the username and verifying the fingerprint. The fingerprint will have to be communicated out of band. Once the first point is settled, I'll update [this PR](https://github.com/interledger/five-bells-shared/pull/50) and then follow that with updating the strategy here

clark800: > using the fingerprint itself as the username

Would this mean that other authentication methods would be removed? Or some usernames are fingerprints and some aren't?

> the Common Name as the username

Do you mean putting the username in the common name field? If so, we should update the spec to restrict usernames to 64 characters: http://security.stackexchange.com/questions/40026/openssl-x509-whats-the-significance-of-cn-common-name

lumberj: > Would this mean that other authentication methods would be removed?

I think we could only use one authentication method at a time. Is there a reason to use multiple?

> Do you mean putting the username in the common name field? 

Yes, but we may instead use the fingerprint as the username and the common name won't be used for authentication.

clark800: > I think we could only use one authentication method at a time.

I don't understand. The username is the same regardless of authentication method. If it's the fingerprint, then what happens to a user who doesn't use client certs?

lumberj: Sorry, I misunderstood Stefan's comments earlier. The option we'll go with is to store a fingerprint alongside the username. When a client authenticates with a certificate, we look them up by fingerprint.

lumberj: I'll make the changes tomorrow here and in `shared`

src/services/config.js
```diff
@@ -30,5 +30,6 @@ if (admin_pass) {
 
 config.auth = {
   basic_enabled: Config.castBool(config.getEnv('AUTH_BASIC_ENABLED'), true),
-  http_signature_enabled: Config.castBool(config.getEnv('AUTH_HTTP_SIGNATURE_ENABLED'), true)
+  http_signature_enabled: Config.castBool(config.getEnv('AUTH_HTTP_SIGNATURE_ENABLED'), true),
+  client_certificates_enabled: Config.castBool(config.getEnv('CLIENT_CERTS_ENABLED'), true)
```
sentientwaffle: Since it doesn't look like a default .pfx is being bundled, shouldn't `client_certificates_enabled` be disabled by default?

sentientwaffle: Oops, nevermind, that's for https.

lumberj: This is a good point of discussion. To use client-cert auth, https must be enabled. If https is enabled, we need a way to configure tls, passing in the certificate options. This should probably exist in five-bells-shared, where we can have a common way to specify tls certificates and keys. I can make a separate PR to do that.

src/models/db/account.js
```diff
@@ -109,7 +116,13 @@ PersistentModelMixin(Account, sequelize, {
   password: Sequelize.STRING,
   public_key: Sequelize.TEXT,
   is_admin: Sequelize.BOOLEAN,
-  is_disabled: Sequelize.BOOLEAN
-})
+  is_disabled: Sequelize.BOOLEAN,
+  fingerprint: Sequelize.STRING
```
clark800: I think you also have to add this to the schema in five-bells-shared and add a migration

lumberj: Thanks. I was wondering about that

lumberj: What schema are you referring to?

clark800: https://github.com/interledger/five-bells-shared/blob/master/schemas/Account.json

lumberj: Added the [schema](https://github.com/interledger/five-bells-shared/pull/50) and migration to this PR

src/services/auth.js
```diff
@@ -48,6 +49,20 @@ passport.use(new HTTPSignatureStrategy(
...
   }))
 
+passport.use(new ClientCertStrategy((fingerprint, info, done) => {
+  if (!config.auth.client_certificates_enabled) {
+    return done(new UnauthorizedError('Unsupported authentication method'))
+  }
+
+  Account.findByFingerprint(fingerprint)
+    .then(function (userObj) {
+      if (!userObj || userObj.is_disabled || userObj.fingerprint !== fingerprint) {
```
clark800: It might be safer to fail authorization if the fingerprint is missing/empty in `userObj`.

lumberj: `userObj.fingerprint !== fingerprint` should cover that case

migrations/20160127110439-account-fingerprint.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
MatthewPhinney: I guess we can get rid of this file, since we aren't supporting migrations.

src/services/auth.js
clark800: Not if `fingerprint` is passed in as an empty string

lumberj: You suggested I add the `!userObj.fingerprint` condition, but this sounds like `!fingerprint`. That would only happen if this implementation were to change without anyone noticing https://github.com/interledger/five-bells-shared/pull/50/files#diff-4d47c0b51e2406b055b5d1c2885e7f9fR96. I will add both cases though, as it just makes the check safer.

clark800: Yeah, it's good practice to be defensive in cases like this that are related to security.

migrations/20160127110439-account-fingerprint.js
lumberj: Have we already removed migrations? If not, it would be best to do that in one go since it looks like there are some other things that would have to change

clark800: Yeah, they are gone now

lumberj:
Closing until TODOs are complete

PR #126, opened 2016-01-27
Remove database migrations

clark800:
We don't need database migrations until we support maintaining a database between major versions. Remove migrations until then makes it easier for us to changes to the database schema such as changing the "password" field to "password_hash", which would be an irreversible migration.


MatthewPhinney:
LGTM

justmoon:
Just FYI: The migrations contain information about indices that isn't captured anywhere else and they've also surfaced bugs that wouldn't have surfaced through any other tests. Such as the issue we ran into where MySQL has a size limit on indices over varchar fields.

You can get all of the benefit with very little with by just having one migration that you update as you go. When you're ready to support upgrades, you just switch to creating new migrations instead of updating that one.

clark800:
Interesting, can you explain more about how that works? My current understanding would be that a single migration would just be a dump of what Sequelize generates.

justmoon:
@clark800 Nevermind - I just reread the Sequelize docs and it turns out that you can define indices in the model options as well.

That means we should be able to just re-add this index:
https://github.com/clark800/five-bells-ledger/commit/c91e23962737d91317d55b4a8208de73a8b32260#diff-07af8044172a44a12437d60ab242a5c2L86

Here:
https://github.com/interledger/five-bells-ledger/blob/master/src/models/db/entry.js#L40
(model options can be passed as a 4th parameter)

Syntax docs here: http://docs.sequelizejs.com/en/latest/docs/models-definition/#indexes

Probably not worth it if we're switching to Knex soon, but there it is.

clark800:
Cool, yeah my next task is to switch to knex though, so hopefully we won't have to worry about sequelize.

justmoon:
Sounds good! :+1:

PR #127, opened 2016-01-28
Support running tests with "mocha" command

clark800:


emschwartz:
What's the logic for changing it to work this way instead of how it was before?

clark800:
Before you couldn't run using the "mocha" command, which means you couldn't do "mocha -b", "mocha -g", etc.

emschwartz:
Getting rid of migrations doesn't sound like a great idea to me but if you do you'd need to change the `circle.yml` file to remove the migrations step from there: https://github.com/interledger/five-bells-ledger/blob/master/circle.yml#L18-L21

clark800:
> This seems hackier than explicitly setting an environment variable

Yeah, it's hacky, but the hack is really having testing-dependent code. I don't think we should assume that the user will always remember to set the environment variable when calling "mocha". Do you have any other ideas? Unfortunately mocha refuses to set "NODE_ENV" automatically or by command line option.

clark800:
> you'd need to change the circle.yml file

Thanks, will do

emschwartz:
I don't think you need to make this change in order to use the command line options for mocha. It works if you add them after `npm run test`

clark800:
> I don't think you need to make this change in order to use the command line options for mocha. It works if you add them after npm run test

Doesn't seem to work for me.

clark800:
Added commands to wipe database in circle.yml, I think this should be ready now.

src/services/config.js
```diff
@@ -8,7 +8,13 @@ config.parseServerConfig()
 config.parseDatabaseConfig()
 config.parseKeyConfig()
 
-if (process.env.NODE_ENV === 'unit') {
+function isRunningUnitTests () {
```
MatthewPhinney: This is a minor point, but I would argue that the tests are not unit tests, since they test the http endpoints, rather than individual methods. With this in mind, I think this function could be renamed to `isRunningTests`. Aside from that, LGTM.

clark800: Yeah, "unit" is somewhat vague, but it's probably better to drop it to avoid potential confusion.

MatthewPhinney:
LGTM

PR #128, opened 2016-01-28
Add /transfers/:id/fulfillment endpoint

clark800:
There will be more refactoring in the future to split the prepare and fulfillment logic into separate methods in the model, but this should be enough to get something working with the new notary API.

### DIFF
--- a/src/lib/app.js
+++ b/src/lib/app.js
@@ -76,6 +76,13 @@ class App {
       models.Transfer.createBodyParser(),
       transfers.putResource)
 
+    router.put('/transfers/:id/fulfillment',
+      passport.authenticate(['basic', 'http-signature', 'anonymous'], {
+        session: false
+      }),
+      models.Transfer.createBodyParser(),
+      transfers.putResource)
+
     router.get('/transfers/:id', transfers.getResource)
     router.get('/transfers/:id/state', transfers.getStateResource)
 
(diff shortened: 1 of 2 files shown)


emschwartz:
What's the point of having this endpoint if it is just going to pass it to the same controller as `PUT /transfers/:id`?

clark800:
This is what the spec requires. The functionality can be split later as described in the description.

emschwartz:
If you say so. A problem I could see arising with this is that if you don't submit the whole transfer exactly as it is in the db it'll give you an invalid modification error

clark800:
> A problem I could see arising with this

Are you referring to this pull request or a pre-existing problem with the code?

MatthewPhinney:
So this is what we need to be compatible with the new notary design? I don't see this endpoint in the ledger tech docs. I do see it, however, in the notary docs. What spec are you referring to?

emschwartz:
With this, if you post the "fulfillment" I'd sort of expect it to just expect and care about the fulfillment. With this endpoint it will look at the whole transfer and return an error if the other fields aren't exactly what the ledger has in the db

clark800:
Actually it sounds like the spec changed or is different than what I had thought: it is only putting the fulfillment part, so this will require more changes.

emschwartz:
Gotcha, that makes more sense to me

clark800:
Replaced by https://github.com/interledger/five-bells-ledger/pull/129

PR #129, opened 2016-01-29
Update case fulfillment to comply with new notary design

MatthewPhinney:


MatthewPhinney:
With this PR, the notary can post a fulfillment to the ledger without needing to include the full transfer object.  One question to address is, should we continue to allow a case to be fulfilled by posting the full transfer with the fulfillment condition? I haven't removed that logic, so, with this PR, there are two ways that a transfer can be executed:
1. `PUT transfers/:id` with the condition fulfillment included in the transfer body
2. `PUT transfers/:id/fulfillment` with the condition fulfillment as the request body
@clark800 @naoitoi let's discuss.

naoitoi:
Thank you for fixing this so quickly, @MatthewPhinney !

Re: the two ways of PUT, I would get rid of the old one (full transfer), to simplify things and avoid future confusion.

clark800:
Agreed, let's make only one way to do it. Doesn't necessarily need to be done in this pull request though.

emschwartz:
It seems like it would be strange if there were a `PUT transfers` endpoint that allowed you to update various transfer fields _except_ the `fulfillment`. What happens if you post a complete transfer that includes the `execution_condition_fulfillment` as well? Would it reject it because you have to submit the fulfillment to a separate endpoint?

src/lib/app.js
```diff
@@ -77,6 +78,18 @@ class App {
       models.Transfer.createBodyParser(),
       transfers.putResource)
 
+    // TODO Need to add body parser for fulfillment object: https://github.com/interledger/five-bells-shared/blob/master/schemas/ConditionFulfillment.json#L32
+    router.put('/transfers/:id/fulfillment',
+      passport.authenticate(['basic', 'http-signature', 'anonymous'], {
+        session: false
+      }), function * (next) {
+        let json = yield parseBody(this)
```
clark800: Could be const or just inlined

clark800:
Is the fulfillment semantically part of the transfer still? I thought it was more like a detached signature now since it is submitted separately. If this is the case, then I think it would make sense to fail if it was attached.

emschwartz:
If we're moving more towards "crypto events" then maybe not. I'm not sure what the latest on that discussion is

clark800:
LGTM :+1: smaller issues can be addressed in the next pull request so we can get the integration tests up faster :smile:

MatthewPhinney:
Do not merge. I want to get the demo up and running with these changes.

MatthewPhinney:
This is ready to be merged. The remainder of this task is tracked under RILP-178.

test/putTransferSpec.js
```diff
@@ -688,7 +701,7 @@ describe('PUT /transfers/:id', function () {
         }))
         .end()
     })
-
+/*
```
MatthewPhinney: The transfer can only be executed when posting to the `/transfers/:id/fulfillment` endpoint. An authorized transfer with no `execution_condition` will no longer execute automatically. Let's discuss, and decide if we want this behavior.

test/putTransferSpec.js
```diff
@@ -842,7 +854,7 @@ describe('PUT /transfers/:id', function () {
...
```
MatthewPhinney: Authorized transfers with no `execution_condition` no longer execute automatically.

justmoon:
> An authorized transfer with no execution_condition will no longer execute automatically. 

What the? So if want to do a standard transfer I now have to not only make two API calls instead of one, but also come up with some dummy condition?!

justmoon:
Sorry for the outburst, but we need to chat. I think there is still some fundamental confusion about this component.

MatthewPhinney:
> Sorry for the outburst, but we need to chat. I think there is still some fundamental confusion about this component.

This was intended to be an open question. I was just marking points in the code that I wanted to reexamine. See my above comment:

> The transfer can only be executed when posting to the /transfers/:id/fulfillment endpoint. An authorized transfer with no execution_condition will no longer execute automatically. Let's discuss, and decide if we want this behavior.

There were a few design issues I wanted to discuss before finishing this issue, so yes, let's chat :smiley:

MatthewPhinney:
@clark800 @justmoon please review. I'd like to merge this and work the remainder of these tasks under a new PR, since this one is becoming bloated.

The implementation is as follows:
Transfers with no `execution_condition` will execute as soon as they are authorized. Transfers with an `execution_condition` (or a `cancellation_condition`) will only execute (or cancel) when a valid fulfillment is `PUT` to `transfers/:id/fulfillment`. This fulfillment must fulfill either the `execution_condition` or the `cancellation_condition` (but not both) or we throw an `UnmetConditionError`. If the `execution_condition` is fulfilled, but the transfer is not prepared, we throw an `InvalidModificationError`. If the `cancellation_condition` is fulfilled, but the transfer is not either `proposed` or `prepared`, we throw an `InvalidModificationError`. Otherwise we execute or cancel the transfer, as appropriate, and persist the fulfillment in the database. The fulfillment may be obtained via the API, along with the transfer object (this behavior hasn't changed).

@emschwartz asked what happens if the user includes the fulfillment when the transfer is `PUT` to `transfers/:id`. At the moment, the fulfillment is ignored: it is neither validated nor persisted. As we continue to decouple fulfillment logic from transfers, we will disallow fulfillments in transfers, persist them in a separate db table, and use a body parser to validate them against the JSON schema. This is work for another PR, and has been specced out in RILP-178. Feel free to review and comment on that task, as well.

src/models/transfers.js
```diff
@@ -207,31 +208,83 @@ function * processStateTransitions (tr, transfer) {
       yield accountBalances.applyDebits()
       updateState(transfer, 'prepared')
     }
+
+    // If the transfer is prepared and has no execution condition, execute immediately
```
clark800: A method called "prepareTransfer" shouldn't be able to execute a transfer

src/models/transfers.js
```diff
@@ -207,31 +208,83 @@ function * processStateTransitions (tr, transfer) {
...
-  }
+    // We don't know if notary is executing or rejecting transfer
+    const isValidExecution = Boolean(transfer.execution_condition && Condition.testFulfillment(transfer.execution_condition, fulfillment))
+    const isValidCancellation = Boolean(transfer.cancellation_condition && Condition.testFulfillment(transfer.cancellation_condition, fulfillment))
 
-  let canRejectState = transfer.state === 'proposed' || transfer.state === 'prepared'
-  if (canRejectState && transfer.cancellation_condition_fulfillment) {
-    if (!transfer.hasValidFulfillment('cancellation')) {
+    // Exactly one condition must be true, so booleans must sum to 1
+    if (isValidCancellation + isValidExecution !== 1) {
```
clark800: `isValidCancellation === isValidExecution` is simpler and I think clear enough to remove the need for comment

src/models/transfers.js
```diff
@@ -207,31 +208,83 @@ function * processStateTransitions (tr, transfer) {
...
 
-    // Remove the expiry countdown
-    transferExpiryMonitor.unwatch(transfer.id)
+    let accountBalances = yield makeAccountBalances(transaction, transfer)
 
-    updateState(transfer, 'executed')
-  }
+    // We don't know if notary is executing or rejecting transfer
+    const isValidExecution = Boolean(transfer.execution_condition && Condition.testFulfillment(transfer.execution_condition, fulfillment))
+    const isValidCancellation = Boolean(transfer.cancellation_condition && Condition.testFulfillment(transfer.cancellation_condition, fulfillment))
```
clark800: How wide is your editor? haha

src/models/transfers.js
MatthewPhinney: Good point. I updated this after naming the method. I think there is some larger refactoring that can be done here, but for now I'll just change the method name.

MatthewPhinney:
@clark800 updated and rebased.

clark800:
LGTM (as long as we finish the refactoring in the next pull request)

justmoon:
@MatthewPhinney 

I won't have time to look at the code line-by-line unfortunately, but this description looks correct to me:

> Transfers with no execution_condition will execute as soon as they are authorized. Transfers with an execution_condition (or a cancellation_condition) will only execute (or cancel) when a valid fulfillment is PUT to transfers/:id/fulfillment. This fulfillment must fulfill either the execution_condition or the cancellation_condition (but not both) or we throw an UnmetConditionError. If the execution_condition is fulfilled, but the transfer is not prepared, we throw an InvalidModificationError. If the cancellation_condition is fulfilled, but the transfer is not either proposed or prepared, we throw an InvalidModificationError. Otherwise we execute or cancel the transfer, as appropriate, and persist the fulfillment in the database. The fulfillment may be obtained via the API, along with the transfer object (this behavior hasn't changed).

Regarding fulfillments - the API is REST, which means that changing the state of a Transfer should be possible by PUTting a new version of that resource. So the old behavior of PUTting a transfer with execution/cancellation_condition_fulfillment should still be supported. I don't know if it is or not in this version of the PR. When we switch to proper crypto-events then the fulfillments might be separate objects, but we haven't spec'd out or signed off on that API yet, so for now it is still part of the transfer and should be handled as such. It also has the nice benefit of maintaining backward compatibility so we don't break the connector submitting fulfillments in universal mode.

And just for background:

> This fulfillment must fulfill either the execution_condition or the cancellation_condition **(but not both)**

(Emphasis added.) When the new crypto-condition spec is finished, the condition will be just a hash, so it would be trivial to enforce a rule that for any given transfer the two conditions cannot be the same. Then any given fulfillment could never fulfill both and that weird edge case goes away.

PR #130, opened 2016-01-30
Split database layer from subscriptions and transfers models

clark800:
This completes the method extraction for the models in this repo.

Remaining work in this refactoring:
1. Use POJOs in interface between db methods and models
2. Use db methods in "lib"


MatthewPhinney:
LGTM

PR #131, opened 2016-02-01
Remove obsolete comments

clark800:
The static verification is done by the JSON schema validator, which is invoked by the "createBodyParser" middleware.

The execution condition fulfillment is checked by `hasValidFulfillment` in `processStateTransitions`.

Is there anything else that should be done?


MatthewPhinney:
LGTM

PR #133, opened 2016-02-03
Separate fulfillment logic from transfers (RILP-178)

MatthewPhinney:


emschwartz:
Please add tests with this

MatthewPhinney:
@emschwartz what kinds of tests did you have in mind? I could send a request with an incorrect JSON payload to the endpoint, but this amounts to testing the JSON schema/validator.  I would argue that such a test belongs in `five-bells-shared` (incidentally, it doesn't look like the validator/schema is tested anywhere). I could also unit test the `createBodyParser` method, but, again, this is inherited from `five-bells-shared`, so I would argue that a test belongs there.

I think that there are some useful tests I can add to this endpoint (a valid fulfillment for an unauthorized transfer should not execute the transfer), but I'm not tracking what specifically should be added to test the validator. What do you think?

src/models/db/conditionFulfillment.js
```diff
@@ -0,0 +1,31 @@
...
+class ConditionFulfillment extends Model {
+  /*
+   * This will be expanded when we fully decouple fulfillments from transfers
+   */
+}
+
+PersistentModelMixin(ConditionFulfillment, sequelize, {
+  id: { type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true },
+  transfer_id: Sequelize.UUID,
+  type: Sequelize.ENUM('execute', 'cancel'),
```
clark800: So this is something that gets stored in the database, but is not supposed to be part of the request? Will it cause any problems if the client puts it in the request?

emschwartz:
That's fair. Adding tests to `shared` would be useful. For this it might be useful to add tests like the one you mention about valid fulfillments for an unauthorized or expired transfer

src/models/db/conditionFulfillment.js
MatthewPhinney: So this may be a little counterintuitive, but this commit is only meant to add body parsing middleware to the `/fulfillments` endpoint (so that we validate the JSON being passed). I added this db model to be consistent with the other endpoints that use this middleware, and because we will probably use it when we fully separate `fulfillment` from `transfer`. If the user passes a fulfillment as part of a transfer, nothing happens (the schema still allows it, but the business logic doesn't use it), but if they put a fulfillment to the `/fulfillments` endpoint, we persist it in the `transfers` table. In other words, the frontend logic is _mostly_ decoupled, but the backend logic is not yet decoupled. I realize this is not ideal, but I can't progress further until we decide:
1. Do we want to fully decouple these components?
2. Do we need to support backwards compatibility?

clark800: > Do we want to fully decouple these components?

Yes

> Do we need to support backwards compatibility?

No

I'd like to also get @justmoon's feedback on this, but let's not block on that since it generally takes a while. We can always revert changes if necessary.

src/models/db/conditionFulfillment.js
```diff
@@ -0,0 +1,80 @@
...
+const InvalidBodyError = require('five-bells-shared/errors/invalid-body-error')
+
+class ConditionFulfillment extends Model {
+
+  // @override
+  static createBodyParser () {
+    const Self = this
+
+    return function * (next) {
+      let json = yield parseBody(this)
```
clark800: const

src/models/db/conditionFulfillment.js
```diff
@@ -0,0 +1,80 @@
...
+      const validationResult = Self.validateExternal(json)
+      if (validationResult.valid !== true) {
+        const message = validationResult.schema
+          ? 'Body did not match schema ' + validationResult.schema
+          : 'Body did not pass validation'
+        throw new InvalidBodyError(message, validationResult.errors)
+      }
+
+      const model = new Self()
```
clark800: Isn't `this` and instance of the class?

src/models/db/conditionFulfillment.js
```diff
@@ -0,0 +1,80 @@
...
+      model.setConditionFulfillment(json)
+      model.setTransferId(this.params.id)
```
clark800: Looks like the HTTP interface is leaking into the models

src/models/transfers.js
```diff
@@ -231,24 +233,26 @@ function * fulfillTransfer (transferId, fulfillment) {
...
+      transfer.cancellation_condition.type === fulfillment.getData().condition_fulfillment.type &&
+      Condition.testFulfillment(transfer.cancellation_condition, fulfillment.getData().condition_fulfillment))
 
     if (isValidCancellation === isValidExecution) {
       throw new UnmetConditionError('ConditionFulfillment failed')
     }
 
-    const canExecute = transfer.state === 'prepared'
-    const canCancel = transfer.state === 'proposed' || transfer.state === 'prepared'
+    const canExecute = transfer.state === 'prepared' || transfer.state === 'executed'
```
clark800: Can you explain this change?

src/models/db/conditionFulfillment.js
MatthewPhinney: Yeah, I actually wasn't sure the best way to handle this. We need to validate the payload against the fulfillment schema, but then we need to attach the `transfer_id` (HTTP param) to the payload, since that is how it's stored in the db. What do you think the best way to do this is?

src/models/transfers.js
MatthewPhinney: So the idea here is that if someone sends a fulfillment for an executed transfer, it should return a 200 code, rather than throwing an error (as it was doing before).

sentientwaffle: Won't this change would allow double-crediting+debiting of accounts?

clark800: What's the reason for that change?

src/models/transfers.js
```diff
@@ -231,24 +233,26 @@ function * fulfillTransfer (transferId, fulfillment) {
 
     // We don't know if notary is executing or rejecting transfer
     const isValidExecution = Boolean(transfer.execution_condition &&
-      Condition.testFulfillment(transfer.execution_condition, fulfillment))
+      transfer.execution_condition.type === fulfillment.getData().condition_fulfillment.type &&
+      Condition.testFulfillment(transfer.execution_condition, fulfillment.getData().condition_fulfillment))
```
sentientwaffle: I think it would be cleaner to test the `condition.type` and `fulfillment.type` from inside `Condition.testFulfillment`

clark800: Does the fulfillment really need a `type`? It seems redundant.

src/models/db/conditionFulfillment.js
clark800: All the HTTP-related stuff (including "next", params, body) has to go in the controller. This entire method might belong in the controller.

src/lib/app.js
```diff
@@ -78,15 +77,16 @@ class App {
...
       passport.authenticate(['basic', 'http-signature', 'anonymous'], {
         session: false
-      }), function * (next) {
-        this.body = yield parseBody(this)
-        yield next
-      },
+      }), models.ConditionFulfillment.createBodyParser(),
       transfers.putFulfillment)
+    router.get('/transfers/:id/fulfillment',
+      passport.authenticate(['basic', 'http-signature', 'anonymous'], {
```
clark800: I think we should only allow "anonymous" mode when necessary. It's not necessary here is it? What about for putting a fulfillment?

src/models/transfers.js
```diff
@@ -231,24 +233,26 @@ function * fulfillTransfer (transferId, fulfillment) {
...
+    const canCancel = transfer.state === 'proposed' || transfer.state === 'prepared' || transfer.state === 'rejected'
 
     if (isValidExecution) {
       if (!canExecute) {
         throw new InvalidModificationError('Transfers in state ' + transfer.state + ' may not be executed')
       }
       yield accountBalances.applyCredits()
-      transfer.execution_condition_fulfillment = fulfillment
       updateState(transfer, 'executed')
+      existingFulfillment = yield fulfillments.upsertFulfillment(fulfillment, {transaction})
```
clark800: The name `existingFulfillment` should be a fulfillment, but actually it's a boolean.

src/models/transfers.js
MatthewPhinney: @sentientwaffle I thought this case had been tested, but doesn't look like it was. Nice catch, thanks. Do we want to throw an error, or allow the user to put the fulfillment many times? If I am `PUT`ting an object, I would expect to be able to be able to do so many times over without receiving an error.

clark800: > If I am PUTting an object, I would expect to be able to be able to do so many times over without receiving an error.

Ah, that's right. You have to allow this to obey the idempotent property of PUT, unless you want to convert to POST.

test/fulfillmentSpec.js
```diff
@@ -0,0 +1,201 @@
...
+  })
+
+  it('should return a fulfillment', function *() {
+    const transfer = this.preparedTransfer
+
+    yield this.request()
+      .put(transfer.id + '/fulfillment')
+      .send(this.executionConditionFulfillment)
+      .expect(201)
+      .expect(_.assign({}, this.executionConditionFulfillment))
```
clark800: What is the `assign` for?

src/models/transfers.js
clark800: Does the fact that PUTing a fulfillment will fail prior to the transfer being created imply a violation of idempotence?

MatthewPhinney: Yeah, it seems like it. Should we convert this to a POST? That would also allow us to easily deal with the double spend problem.

clark800: I can't seem to find a definition that clearly answers this question. I don't see a problem with POST if that makes it easier.

src/models/transfers.js
MatthewPhinney: I'm not opposed to re-working the logic in the condition library and making corresponding changes to this code, but until we do that, we need to check that the fulfillment type matches the condition type, or `five-bells-condition` may throw an error. I also think re-working the checking of conditions is a task for another (hopefully much smaller) PR. We can open a ticket for that, if you'd like.

clark800: Sounds good

src/models/transfers.js
MatthewPhinney: Talking with @naoitoi, we feel that changing `PUT` to `POST` is a larger design decision that we should discuss as a team in more detail. Maybe for now I can revert so we throw an error when the user tries to fulfill a transfer multiple times? While this technically doesn't work for `PUT`, it solves the current problem, where we execute a transfer multiple times. That way we can discuss how we want the API to behave in more detail without holding up this PR.

clark800: Well we know for sure that is not a valid solution. The POST option is at least possibly a valid solution, so I think that is a better option if you want to get this merged. We can still discuss it and change it if necessary.

clark800: To clarify, I mean non-idempotent PUT is not a valid solution. If you make PUT idempotent then that is valid.

MatthewPhinney: I don't see how to make PUT idempotent, since it will throw an error if the transfer is not yet created (did we agree this is not idempotent behavior?). I think the behavior I'd like to support is more consistent with POST rather than PUT. But it is an API breaking change. What do you think @naoitoi?

naoitoi: @MatthewPhinney @clark800 I understand the concern.  Changing the API to POST is fine by me.

MatthewPhinney:
As discussed, I changed PUT to POST for the fulfillment endpoint. I also added checks on the account balance after each transfer is executed (or cancelled) to make sure the funds end up in the appropriate account.

src/models/db/conditionFulfillment.js
```diff
@@ -0,0 +1,61 @@
...
+
+  static findByTransfer (transferId, options) {
+    return ConditionFulfillment.findOne({
+      where: {transfer_id: transferId},
+      transaction: options && options.transaction
+    })
+  }
+
+  setTransferId (transferId) {
+    super.setData({
```
sentientwaffle: Why is this `super.setData` instead of `this.setData`?

src/models/db/conditionFulfillments.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+function * _upsertFulfillment (fulfillment, options) {
+  assert(options.transaction)
+  // SQLite's implementation of upsert does not tell you whether it created the
+  // row or whether it already existed. Since we need to know to return the
+  // correct HTTP status code we unfortunately have to do this in two steps.
+  const existingFulfillment = yield ConditionFulfillment.findByTransfer(fulfillment.getData().transfer_id, options)
+  if (existingFulfillment) {
+    ConditionFulfillment.upsert(existingFulfillment, options)
```
sentientwaffle: Why upsert the `existingFulfillment`?

src/models/db/conditionFulfillment.js
MatthewPhinney: `setData` is a method in the super class, so I was using `super` for clarity.

src/controllers/transfers.js
```diff
@@ -158,13 +158,22 @@ function * putFulfillment () {
   const id = this.params.id
   requestUtil.validateUriParameter('id', id, 'Uuid')
   const fulfillment = this.body
-  this.body = yield model.fulfillTransfer(id, fulfillment)
-  this.status = 200
+  fulfillment.setTransferId(id)
```
clark800: Rename this function to postFulfillment?

clark800:
LGTM aside from comment

PR #137, opened 2016-02-05
Update config module

lumberj:
Update to use new config
https://github.com/interledger/five-bells-shared/pull/63/

### DIFF
--- a/src/services/config.js
+++ b/src/services/config.js
@@ -1,40 +1,38 @@
 'use strict'
 
 const Config = require('five-bells-shared').Config
-
-const config = module.exports = new Config('ledger')
-
-config.parseServerConfig()
-config.parseDatabaseConfig()
-config.parseKeyConfig()
+const envPrefix = 'ledger'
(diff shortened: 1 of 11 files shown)


src/services/config.js
```diff
@@ -1,40 +1,38 @@
...
-  config.server.public_host = 'localhost'
-  config.server.port = 61337
-  config.server.public_port = 80
-  // We use a different config parameter for unit tests, because typically one
-  // wouldn't want to use their production or even dev databases for unit tests
-  // and it'd be far to easy to do that accidentally by running npm test.
-  config.db.uri = process.env.LEDGER_UNIT_DB_URI || 'sqlite://'
-  config.updateDerivativeServerConfig()
-}
+let localConfig = {}
```
clark800: this is still const because the symbol is never reassigned

lumberj: Good call

clark800:
LGTM

PR #143, opened 2016-02-12
Change fulfillment endpoint to PUT

MatthewPhinney:


src/models/transfers.js
```diff
@@ -253,16 +256,18 @@ function * fulfillTransfer (transferId, fulfillment) {
...
-      if (transfer.state !== 'prepared') {
+      if (!_.includes(validExecutionStates, transfer.state)) {
         throw new InvalidModificationError('Transfers in state ' +
           transfer.state + ' may not be executed')
       }
-      yield accountBalances.applyCredits()
+      if (transfer.state === 'prepared') {
+        yield accountBalances.applyCredits()
+      }
       updateState(transfer, 'executed')
```
clark800: If the transfer was already executed, shouldn't it be a no-op?

MatthewPhinney: What are you thinking?
```
if (transfer.state === 'executed') {
  return {
    fulfillment: fulfillment.getDataExternal(),
    existed: true
  }
}
```
Something like that?

clark800: To be safe, we might need to look up the existing fulfillment because for example there may be multiple fulfillments that satisfy the condition and we should probably be returning the first fulfillment since the second one has no actual effect. So after `validateConditionFulfillment`, if the transfer state is `execution`, just look up the existing fulfillment and return like the code in your comment.

MatthewPhinney:
@clark800 let me know what you think. If it looks good, I'll squash the commits and increment the version.

src/models/transfers.js
```diff
@@ -250,28 +270,23 @@ function * fulfillTransfer (transferId, fulfillment) {
     }
 
     const isExecution = validateConditionFulfillment(transfer, fulfillment)
-    const accountBalances = yield makeAccountBalances(transaction, transfer)
+
+    if (isExecution && transfer.state === 'executed' || !isExecution && transfer.state === 'rejected') {
+      return (yield fulfillments.getFulfillment(transferId)).getDataExternal()
```
clark800: This looks like an inconsistent return type

clark800: Nevermind, this is a return from a nested function

MatthewPhinney: I can't think of a clean way to make this clearer, so I'm inclined to leave as is for now. It's not perfect, but I'd prefer to revisit later (if we decide it needs refactoring) than be held up now.

clark800:
LGTM

PR #145, opened 2016-02-16
BREAKING: Specify "password" instead of "password_hash" in account PUT

clark800:


justmoon:
API LGTM

(Did not review implementation in detail.)

src/models/db/account.js
```diff
@@ -10,6 +11,24 @@ const EntryGroup = require('./entry-group').EntryGroup
 const Sequelize = require('sequelize')
 const sequelize = require('../../services/db')
 
+function hashPassword (password) {
+  // cache hashes of passwords used in tests to speed up tests 4X
```
MatthewPhinney: I'm all for speeding up tests, but it seems weird to have a hardcoded list of test passwords in this method. Can we pull the list from a file under `/test/data`?

naoitoi:
Could you provide more context?  Why is this change necessary?  Is the API change documented in the spec or API doc?  What would this break?

MatthewPhinney:
Do we need to update any other components (connector, notary) for this change?

src/models/db/account.js
clark800: No, production code does not have access to test code. We could move it to a separate file in the production code, but I'm not sure if it's worth it since it's pretty small.

clark800:
The motivation for this change is to decouple account creation from bcrypt so that clients creating accounts do not have to hash the passwords. I do not think this is documented. This will break any components that create accounts on the ledger, though I'm not sure which ones that is currently.

justmoon:
Rationale for the change:

How an application hashes its passwords is an implementation detail and should not be exposed in the API.
* The password is still passed in cleartext with every authentication, so we gain no security by passing a hash here.
* By baking the hashing algorithm into the API, we make gradual upgrades to more secure hashing algorithms difficult/impossible since all clients have to upgrade.
* This is also unnecessarily adding a fairly heavy dependency (bcrypt) to any client.
* Having the algorithm baked into the API fixes the algorithm parameters, so a ledger with higher security requirements wouldn't be able to choose bcrypt with more rounds or a more secure algorithm than bcrypt. (Technically they could layer another instance of bcrypt on top, but's that's just getting silly.)

What has to be updated:

* five-bells-connector (autofund feature)
* five-bells-demo (account setup)
* ILP server (account setup in tests) - assumed
* five-bells-ledger-uri (account creation)

The change to clients is very simple - just take out the client-side bcrypt logic and pass the password directly.

MatthewPhinney:
LGTM

PR #146, opened 2016-02-16
Include transfer fulfillment in notification

MatthewPhinney:


naoitoi:
Could you provide more context?  Why is this change necessary?  Does this change API?  If so, is it documented in spec or API doc?

MatthewPhinney:
@naoitoi
https://github.com/interledger/five-bells-connector/issues/79
https://ripplelabs.atlassian.net/browse/RILP-215
As far as I know, there is no documentation. This change should fix universal mode (which is currently broken). It should not change atomic mode.

emschwartz:
The reason for this change is that in universal mode the connector needs to get the fulfillment from the notification so it can pass it to the previous ledger and execute the source_transfers. The fulfillment changes made in the last couple of commits broke this functionality

emschwartz:
@MatthewPhinney Note that this requires a corresponding change in https://github.com/interledger/five-bells-shared/blob/master/schemas/Notification.json

What do you think about adding a test that checks whether the notification returned is valid according to the schema?

justmoon:
> Could you provide more context? Why is this change necessary? Does this change API? If so, is it documented in spec or API doc?

lol, I wonder if @naoitoi wrote a bot that just posts this message for all pull requests that have a zero length description. Doesn't matter, automatic or manual, I approve! Let's up our pull request game.

Rationale:

* In Universal, the connector must pass the fulfillment from the payout ledger (right-hand ledger) to the payin ledger (left-hand side) in order to recoup its investment.
* Previously, the transfer itself contained the fulfillment, but this is no longer the case.
* To manually fetch the fulfillment would be a whole extra roundtrip, adding latency to all Universal mode transactions.
* Simplest fix: Put the fulfillment into the notification.

Design notes:

First some background -

It is actually a very common case in REST APIs that you are passing a resource and you know that your client is going to need other related resources. Imagine a web browser fetching google.com. Google knows that in a moment, that same browser is also going to request the Google logo image that is on that page. In a REST API it could be something like requesting a list of posts, but to render then you'll need to request the authors also, so you can render the posts with author name, avatar etc.

We discussed this at length a while ago and considered several options:

* [JSON-LD](http://json-ld.org/)
* [JSON+HAL _embedded](http://stateless.co/hal_specification.html)
* [HTTP2 server push](https://www.igvita.com/2013/06/12/innovating-with-http-2.0-server-push/)

JSON-LD does serve the use case, but was quite difficult to understand and even after putting considerable time into it we still weren't sure if our implementation was truly compliant with the standard.

JSON+HAL is very simple, but quite invasive and didn't feel like a very elegant way of doing, more like a hack.

Because this is purely a performance optimization (semantically it is totally fine to make a round-trip for each resource), we like the idea of solving the issue below the API layer. HTTP2 push is perfect for our use case. If somebody requests a transfer `GET /transfers/123` and we think they'll also need a related account `GET /accounts/bob`, we can HTTP2 push that resource to them. So then when they request that resource, they already have it cached client-side. No extra JSON format to understand, no extra parsing, but still 100% of the performance gain.



Now let's get back to this pull request. HTTP2 server push doesn't work here. It would be the client pushing. So that's why we're adding a `related_resources` field to notifications that doesn't appear anywhere else in the API.

In short:

If you're worried about the performance of fetching related resources and you are
* A server responding to a request --> Use HTTP2 server push (we should add HTTP2 support to FB components in the near-ish future.)
* A server posting a notification --> Use `related_resources`

(I'll add all this to the [Design Rationale](https://docs.google.com/document/d/1dO6PKxWXHwLoO4wq727aUjt8zS428cHUyT8W69uqdCc/edit) doc.)

> What do you think about adding a test that checks whether the notification returned is valid according to the schema?

:+1:

MatthewPhinney:
Let's track checking against the schema using RILP-114. I have tests that check the structure of the JSON here: https://github.com/interledger/five-bells-ledger/pull/146/files#diff-a94639f5e1b76fc887070e535f6521a3R366, so we should be OK, until RILP-114 is completed.

clark800:
LGTM

PR #152, opened 2016-02-17
Server-side Client Certificate Authentication

lumberj:
Server-side Client Certificate Authentication

Uses https://github.com/ripple/passport-client-certificate

Working example:

Server
```diff
   _makeRouter () {
     const router = new Router()
-    router.get('/health', health.getResource)
+
+    router.get('/health',
+        passport.authenticate(['client-cert']),
+        health.getResource)

```

Client
```js
var fs = require('fs');
var https = require('https');

var options = {
  hostname: 'localhost',
  port: 3000,
  path: '/health',
  method: 'GET',
  key: fs.readFileSync('client1-key.pem'),
  cert: fs.readFileSync('client1-crt.pem'),
  ca: fs.readFileSync('ca-crt.pem') // Trust the test servers self-signed cert
};

var req = https.request(options, function(res) {
  res.on('data', function(data) {
    process.stdout.write(data);
  });
});
req.end();
```

```shell
$ LEDGER_TLS_KEY=server-key.pem \
  LEDGER_TLS_CERTIFICATE=server-crt.pem \
  LEDGER_TLS_CA=ca-crt.pem \
  LEDGER_PUBLIC_HTTPS=true LEDGER_DB_SYNC=1 \
  LEDGER_DB_URI=sqlite://:memory: \
  npm start
```

Certificates for this example can be found at:
https://github.com/ripple/passport-client-certificate/tree/master/test/data

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "moment": "^2.10.2",
     "nodemon": "^1.3.5",
     "passport-anonymous": "^1.0.1",
+    "passport-client-certificate": "^0.1.1",
     "passport-http": "^0.3.0",
     "passport-http-signature": "^1.0.0",
     "priorityqueuejs": "^1.0.0",
--- a/src/lib/app.js
+++ b/src/lib/app.js
@@ -17,6 +17,7 @@ const subscriptions = require('../controllers/subscriptions')
 const notifications = require('../controllers/notifications')
 const models = require('../models/db')
 const seedDB = require('./seed-db')
+const fs = require('fs')
 
 // Configure passport
 require('../services/auth')
@@ -59,7 +60,32 @@ class App {
     yield this.db.init()
     yield seedDB(this.config)
 
-    this.koa.listen(this.config.getIn(['server', 'port']))
+    if (this.config.getIn(['server', 'secure'])) {
+      const https = require('https')
+      const tls = this.config.get('tls').toJS()
+
+      const options = {
+        port: this.config.getIn(['server', 'port']),
+        host: this.config.getIn(['server', 'bind_ip']),
+        key: fs.readFileSync(tls.key),
+        cert: fs.readFileSync(tls.cert),
+        ca: tls.ca && fs.readFileSync(tls.ca),
+        crl: tls.crl && fs.readFileSync(tls.crl),
+        requestCert: this.config.getIn(['auth', 'client_certificates_enabled']),
+
+        // Certificates are checked in the passport-client-cert middleware
+        // Authorization check is disabled here to allow clients to connect
+        // to some endpoints without presenting client certificates, or using a
+        // different authentication method (e.g., Basic Auth)
+        rejectUnauthorized: false
+      }
+
+      https.createServer(
+        options, this.koa.callback()).listen(this.config.getIn(['server', 'port']))
+    } else {
+      this.koa.listen(this.config.getIn(['server', 'port']))
+    }
+
     this.log.info('ledger listening on ' +
       this.config.getIn(['server', 'bind_ip']) + ':' +
       this.config.getIn(['server', 'port']))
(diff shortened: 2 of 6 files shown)


migrations/20160127110439-account-fingerprint.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    queryInterface.addColumn('Accounts', 'fingerprint', Sequelize.STRING)
```
lumberj: I think I should be adding the index in the migration as well

src/services/config.js
```diff
@@ -20,11 +20,6 @@ if (admin_pass) {
   }
 }
 
-localConfig.auth = {
```
lumberj: Moved to five-bells-shared

lumberj:
cc: @justmoon and @sentientwaffle. At the moment this is failing because `passport-client-cert` is not yet included, as per the commit message. This is tested locally and everything works.

migrations/20160127110439-account-fingerprint.js
lumberj: Actually, I should take this out as we've removed migrations

clark800:
Can you rebase?

package.json
```diff
@@ -31,7 +31,7 @@
     "co-request": "^1.0.0",
     "deep-diff": "^0.3.0",
     "five-bells-condition": "~2.0.0",
-    "five-bells-shared": "~9.1.2",
+    "five-bells-shared": "~9.2.1",
```
MatthewPhinney: This was also updated here: https://github.com/interledger/five-bells-ledger/commit/d682b3c2480a981939f46db0afb06218251a8d7e, which is probably causing the merge conflicts.

src/services/auth.js
```diff
@@ -54,6 +55,22 @@ passport.use(new HTTPSignatureStrategy(
...
+passport.use(new ClientCertStrategy((certificate, done) => {
+  if (!config.getIn(['auth', 'client_certificates_enabled'])) {
+    return done(new UnauthorizedError('Unsupported authentication method'))
+  }
+
+  const fingerprint = certificate.fingerprint.toUpperCase()
+  Account.findByFingerprint(fingerprint)
+    .then(function (userObj) {
+      if (!userObj || userObj.is_disabled || !userObj.fingerprint ||
+          !fingerprint || userObj.fingerprint !== fingerprint) {
```
MatthewPhinney: What's the rational behind checking `!fingerprint` here?

src/lib/app.js
```diff
@@ -59,7 +60,33 @@ class App {
     yield this.db.init()
     yield seedDB(this.config)
 
-    this.koa.listen(this.config.getIn(['server', 'port']))
+    if (this.config.getIn(['server', 'secure'])) {
+      const https = require('https')
+      const tls = this.config.get('tls').toJS()
+
+      // https://nodejs.org/api/ssl.html#ssl_ssl_createserver_options_secureconnectionlistener
```
naoitoi: I would elaborate what this URL is.

naoitoi:
LGTM.

src/models/db/account.js
```diff
@@ -54,6 +54,7 @@ class Account extends Model {
     delete data.primary
     delete data.password_hash
     delete data.public_key
+    delete data.fingerprint
```
clark800: What's the reason for deleting?

src/services/auth.js
lumberj: Good catch, should have deleted that

src/lib/app.js
lumberj: I think I'll remove it because the link can go stale or change

src/models/db/account.js
lumberj: II think for the same reason the public_key is not shared

lumberj: The fingerprint is a hash of the contents of the client certificate. I don't think there's any case where this needs to be communicated back to a client

MatthewPhinney:
LGTM

PR #153, opened 2016-02-18
Update npm-shrinkwrap.json

justmoon:
NPM is still installing old versions of five-bells-shared and other dependencies, because there is an outdated `npm-shrinkwrap.json` in the root of the project.


clark800:
Should we bump the version to make this a release? I'm trying to do this right now and getting `npm test` failures.

clark800:
Used https://github.com/interledger/five-bells-ledger/pull/155 instead because I needed the version bump to get integration tests working

PR #154, opened 2016-02-18
Add schema validation to account tests

MatthewPhinney:


MatthewPhinney:
Sample error message:
```
 6) Accounts GET /accounts/:uuid should strip out the password field:
     Error: Account schema validation error: {"message":"Additional properties not allowed","params":{"key":"foo"},"code":303,"dataPath":"/foo","schemaPath":"/additionalProperties","subErrors":null,"name":"ValidationError"}
      at test/accountSpec.js:26:13
      at Array.forEach (native)
      at validAccounts (test/accountSpec.js:22:12)
      at Test._assertFunction (node_modules/supertest/lib/test.js:247:11)
      at Test.assert (node_modules/supertest/lib/test.js:148:18)
      at assert (node_modules/supertest/lib/test.js:127:12)
      at node_modules/supertest/lib/test.js:124:5
      at Test.Request.callback (node_modules/supertest/node_modules/superagent/lib/node/index.js:819:12)
      at IncomingMessage.<anonymous> (node_modules/supertest/node_modules/superagent/lib/node/index.js:1049:12)
      at endReadableNT (_stream_readable.js:905:12)
```

test/accountSpec.js
```diff
@@ -11,10 +11,29 @@ const dbHelper = require('./helpers/db')
...
 const logHelper = require('five-bells-shared/testHelpers/log')
 
+const accountValidator = require('../src/services/validator').create('Account')
+const transferValidator = require('../src/services/validator').create('Transfer')
 const publicKey = fs.readFileSync(__dirname + '/data/public.pem', 'utf8')
 
 const START_DATE = 1434412800000 // June 16, 2015 00:00:00 GMT
 
+function validAccounts (res) {
+  const accounts = Array.isArray(res.body) ? res.body : [res.body]
```
clark800: isn't it always an array?

MatthewPhinney: It's actually only an array on `.get('/accounts')`. In other cases, it's a single account.

test/accountSpec.js
```diff
@@ -11,10 +11,29 @@ const dbHelper = require('./helpers/db')
...
+  accounts.forEach(account => {
+    const validatorResult = accountValidator(account)
+    if (!validatorResult.valid) {
+      throw new Error('Account schema validation error: ' + JSON.stringify(_.omit(validatorResult.errors[0], ['stack'])))
+    }
+  })
+}
+
+function validTransfer (res) {
```
clark800: I think the naming convention we've been using is `validateTransfer`

test/accountSpec.js
clark800: Oh, how about using two functions `validateAccount` and `validateAccounts`?

test/accountSpec.js
```diff
@@ -11,10 +11,32 @@ const dbHelper = require('./helpers/db')
...
+}
+
+function validAccount (res) {
+  const validatorResult = accountValidator(res.body)
+  if (!validatorResult.valid) {
+    throw new Error('Account schema validation error: ' + JSON.stringify(_.omit(validatorResult.errors[0], ['stack'])))
+  }
+}
+
+function validTransfer (res) {
```
MatthewPhinney: Let's discuss the naming convention here. I agree that, as a function, this should most likely be called `validateTransfer`, but with the BDD chai `expect`, it kind of makes sense to call it `validTransfer` (we expect a valid transfer to be returned from the endpoint). I'm open to suggestions.

clark800: I haven't used BDD style much, but I've seen things like `expect(validateTransfer).to.not.throw(Error)`, does that fit in this case?

PR #157, opened 2016-02-18
Add schema validation to account tests

MatthewPhinney:


test/accountSpec.js
```diff
@@ -93,6 +116,7 @@ describe('Accounts', function () {
             connector: 'http://localhost:4321'
           }
         ])
+        .expect(validAccounts)
```
MatthewPhinney: We could do something like
```
.expect(res => {
  expect(validAccounts.bind(validAccounts, res)).to.not.throw(Error)
 })
```
But that's an extra two lines of code each time we want to validate a schema, and in my opinion less readable. I'm still searching for an ideal solution.

clark800:
LGTM

PR #158, opened 2016-02-18
Validate fulfillment responses against JSON schema

MatthewPhinney:


test/fulfillmentSpec.js
```diff
@@ -331,6 +369,7 @@ describe('PUT /fulfillment', function () {
           subscription: subscription.id,
           resource: transferPrepared
         })
+        expect(validateNotification.bind(validateNotification, body)).to.not.throw(Error)
```
clark800: Why is this one different?

MatthewPhinney: This is not a chained expect, so I have to pass the function `validateNotification` along with the value it should accept in order for `expect` to evaluate it appropriately. See here: http://stackoverflow.com/questions/21587122/mocha-chai-expect-to-throw-not-catching-thrown-errors.

clark800:
LGTM

PR #159, opened 2016-02-19
Validate transfer schemas

MatthewPhinney:
This should not be merged until #158 is merged. I need to use the newer version of five-bell-shared, so that I can validate the notifications.

There is a lot of duplicated code between the tests. Once I have added schema validation to each test, I'll consolidate, probably by adding a file with some helper methods, so we don't need the validators in each test file. I'm submitting PRs one by one, so they are easier to review.


clark800:
LGTM

PR #164, opened 2016-02-23
Add schema validation to transfer state receipt tests

MatthewPhinney:
This will fail until https://github.com/interledger/five-bells-shared/pull/88 is merged.

### DIFF
--- a/src/models/transfers.js
+++ b/src/models/transfers.js
@@ -29,6 +29,8 @@ const Condition = require('five-bells-condition').Condition
 
 const validExecutionStates = ['prepared']
 const validCancellationStates = ['prepared', 'proposed']
+const SIGNATURE_TYPE_ED25519 = 'ed25519-sha512'
+const SIGNATURE_TYPE_SHA256 = 'sha256'
 
 function * getTransfer (id) {
   log.debug('fetching transfer ID ' + id)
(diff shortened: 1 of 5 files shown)


src/models/transfers.js
```diff
@@ -29,6 +29,8 @@ const Condition = require('five-bells-condition').Condition
 
 const validExecutionStates = ['prepared']
 const validCancellationStates = ['prepared', 'proposed']
+const SIGNATURE_TYPE_ED25519 = 'ed25519-sha512'
+const SIGNATURE_TYPE_SHA256 = 'sha256'
```
clark800: I think this is going to be confusing to people because "sha256" is a hashing algorithm, which is not a signature scheme. I assume we are using it like a signature here, but does it actually satisfy the technical definition of a digital signature? If so, I'm guessing it has a more accurate name than just the name of the hashing algorithm.

emschwartz: This is one of the papers we referenced in the whitepaper that shows "[One-Way Functions are Necessary and Sufficient for Secure Signatures](http://www.cs.princeton.edu/courses/archive/spr08/cos598D/Rompel.pdf)". Not sure what the best approach is for naming. We might need to just put some documentation somewhere about how hash algorithms can be used for signatures and put links to that anywhere we think people might get confused

clark800: I discussed with @justmoon and we were thinking that it might be more accurate to call these "commitments" rather than "signatures". I'm still not fully caught up with understanding how this implementation works, but does this sound right? https://en.wikipedia.org/wiki/Commitment_scheme

It seems like we aren't using it in the typical way that commitments are used, but cryptographically it seems to be closer to a commitment than a signature.

emschwartz: Sounds alright to me

clark800:
LGTM

PR #165, opened 2016-02-23
Allow TLS client auth for subscription notifications

lumberj:
~~Depends on https://github.com/interledger/five-bells-ledger/pull/166~~

### DIFF
--- a/src/lib/notificationUtils.js
+++ b/src/lib/notificationUtils.js
@@ -1,0 +1,32 @@
+'use strict'
+
+const request = require('co-request')
+const _ = require('lodash')
+const url = require('url')
+
+function isHTTPS (uri) {
+  return url.parse(uri).protocol.match(/https/) !== null
+}
+
+function tlsOptions (target, config) {
+  const tls = config.get('tls')
+  const useTLS = isHTTPS(target) && tls
+  return useTLS ? _.omit({
+    cert: tls.get('cert'),
+    key: tls.get('key'),
+    ca: tls.get('ca'),
+    crl: tls.get('crl')
+  }, _.isUndefined) : {}
+}
+
+function * sendNotification (target, notificationBody, config) {
+  yield* request(target, _.assign({
+    method: 'post',
+    json: true,
+    body: notificationBody
+  }, tlsOptions(target, config)))
+}
+
+module.exports = {
+  sendNotification: sendNotification
+}
--- a/src/lib/notificationWorker.js
+++ b/src/lib/notificationWorker.js
@@ -3,10 +3,10 @@
 const _ = require('lodash')
 const co = require('co')
 const defer = require('co-defer')
-const request = require('co-request')
+const utils = require('./notificationUtils')
 
 class NotificationWorker {
-  constructor (uri, log, Notification, Transfer, Subscription, Fulfillment) {
+  constructor (uri, log, Notification, Transfer, Subscription, Fulfillment, config) {
     this._timeout = null
 
     this.uri = uri
@@ -15,6 +15,7 @@ class NotificationWorker {
     this.Transfer = Transfer
     this.Subscription = Subscription
     this.Fulfillment = Fulfillment
+    this.config = config
 
     this.processingInterval = 1000
     this.initialRetryDelay = 2000
@@ -134,11 +135,8 @@ class NotificationWorker {
       }
     }
     try {
-      const result = yield request(subscription.target, {
-        method: 'post',
-        json: true,
-        body: notificationBody
-      })
+      const result = yield utils.sendNotification(
+        subscription.target, notificationBody, this.config)
       // Success!
       if (result.statusCode < 400) {
         yield notification.destroy()
--- a/src/services/config.js
+++ b/src/services/config.js
@@ -1,33 +1,5 @@
 'use strict'
 
-const Config = require('five-bells-shared').Config
-const envPrefix = 'ledger'
+const loadConfig = require('../lib/config')
 
-function isRunningTests () {
-  return process.argv[0].endsWith('mocha') ||
-    (process.argv.length > 1 && process.argv[0].endsWith('node') &&
-     process.argv[1].endsWith('mocha'))
-}
-
-const localConfig = {}
-const admin_user = Config.getEnv(envPrefix, 'ADMIN_USER') || 'admin'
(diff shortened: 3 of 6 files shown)


src/lib/notificationWorker.js
```diff
@@ -15,6 +15,7 @@ class NotificationWorker {
     this.Transfer = Transfer
     this.Subscription = Subscription
     this.Fulfillment = Fulfillment
+    this.config = require('../services/config')
```
lumberj: We want to avoid requiring stateful modules (anything in src/services) in any of the src/lib modules. There are two other options I considered instead of having a `require` here.

1) Pass the service/config into the constructor
2) Require lib/config in this module, and load a config object for every instance of NotificationWorker.

I think option 2 is nicer, but I want to get hear any opinions on this.

lumberj:
cc: @justmoon @geertweening

lumberj:
Rebased against five-bells-shared v10 updated

src/lib/notificationWorker.js
MatthewPhinney: What is `lib/config`?

lumberj: @MatthewPhinney good point, I got confused with connector. I updated the comment

MatthewPhinney: In that case, I think option 2 makes sense.

lumberj: Just updated with a commit to add that

src/lib/notificationUtils.js
```diff
@@ -0,0 +1,32 @@
...
+  const useTLS = isHTTPS(target) && tls
+  return useTLS ? _.omit({
+    cert: tls.get('cert'),
+    key: tls.get('key'),
+    ca: tls.get('ca'),
+    crl: tls.get('crl')
+  }, _.isUndefined) : {}
+}
+
+function * notificationRequest (target, notificationBody, config) {
```
clark800: I think it might be more clear to call this `sendNotification`

lumberj:
There's currently no unit tests for the `NotificationWorker`, but ideally that should be added after this PR to test the changes. It's indirectly tested in https://github.com/interledger/five-bells-ledger/blob/master/test/subscriptionSpec.js

src/lib/notificationUtils.js
lumberj: Updated

src/services/config.js
```diff
@@ -1,33 +1,5 @@
...
-    ed25519: {
-      secret: 'iMx6i3D3acJPc4aJlK0iT/pkJP3T+Dqte9wg6hXpXEv08CpNQSm1J5AI6n/' +
-        'QVBObeuQWdQVpgRQTAJzLLJJA/Q==',
-      public: '9PAqTUEptSeQCOp/0FQTm3rkFnUFaYEUEwCcyyySQP0='
-    }
-  }
-}
-
-module.exports = Config.loadConfig(envPrefix, localConfig)
+module.exports = loadConfig()
```
clark800: It looks like we're using a mixed paradigm now; `NotificationWorker` calls `loadConfig` directly whereas other parts of the code use this module. I think it would be good to decide on the best method and use it consistently.

lumberj: To be consistent with the existing paradigm, I would add `config` to the constructor of NotificationWorker

clark800: That seems reasonable. The simplest option might be to have the "services/config" module in the lowest architectural layer so that it is accessible anywhere.

lumberj: > The simplest option might be to have the "services/config" module in the lowest architectural layer so that it is accessible anywhere.

Can you elaborate on this?

clark800: You originally said you didn't want to depend on the "services" directory, but what if you just moved the "services/config" file into "lib" or some other directory?

lumberj: I am now passing `config` into the constructor, which is the prevailing paradigm here. If we want to reduce the number of parameters in the constructor method, we can mimic keyword args with an object like in services/app.js

https://github.com/interledger/five-bells-ledger/blob/master/src/services/app.js

src/lib/notificationUtils.js
```diff
@@ -0,0 +1,32 @@
...
+  return useTLS ? _.omit({
+    cert: tls.get('cert'),
+    key: tls.get('key'),
+    ca: tls.get('ca'),
+    crl: tls.get('crl')
+  }, _.isUndefined) : {}
+}
+
+function * sendNotification (target, notificationBody, config) {
+  yield* request(target, _.assign({
```
clark800: Can you explain why this is `yield*` instead of `yield`?

src/lib/notificationUtils.js
```diff
@@ -0,0 +1,32 @@
...
+const url = require('url')
+
+function isHTTPS (uri) {
+  return url.parse(uri).protocol.match(/https/) !== null
+}
+
+function tlsOptions (target, config) {
+  const tls = config.get('tls')
+  const useTLS = isHTTPS(target) && tls
+  return useTLS ? _.omit({
```
MatthewPhinney: Why `omit`? Do we expect any of these values to be undefined if `tls` is defined?

clark800:
LGTM

src/lib/notificationUtils.js
lumberj: It's a generator that's yielding the result of another generator. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*

src/lib/notificationUtils.js
lumberj: ca and crl are optional

src/lib/notificationUtils.js
clark800: Doesn't `yield` do that too? What happens if you change to `yield`?

lumberj: `yield*` is like delegation, when you call this function (`sendNotification`) with yield, it just yields the value from the inner generator function (`request`). I don't think it works correctly if you use `yield`

clark800: I tried changing it to just `yield` and all tests pass. I think we are using `yield` this way in other places in the code. I think you only need `yield*` if the generator yields multiple times and you want to forward all of the values up the call stack.

PR #167, opened 2016-02-23
Add endpoints for ed25519 public key and hash preimage (RILP-95)

MatthewPhinney:


src/controllers/transfers.js
```diff
@@ -61,6 +61,17 @@ function * getStateResource () {
     id.toLowerCase(), signatureType, conditionState)
 }
 
+function getPublicKey () {
+  this.body = model.getPublicKey()
```
MatthewPhinney: Is it bad practice to not use a generator if there is no code in the endpoint that will run asynchronously? The code works as written, but maybe it would be clearer to have all endpoints use a generator. I'm not sure if there are performance trade offs.

src/lib/app.js
```diff
@@ -116,6 +116,8 @@ class App {
 
     router.get('/transfers/:id', transfers.getResource)
     router.get('/transfers/:id/state', transfers.getStateResource)
+    router.get('/transfers/:id/hash_preimage', transfers.getPreimage)
```
MatthewPhinney: The default behaviour for this endpoint is to return the preimage of the transfer's current state, unless the `condition_state` is specified as a query param, in which case we return the hash preimage for the specified state.

src/lib/app.js
```diff
@@ -116,6 +116,8 @@ class App {
...
+    router.get('/transfers/public_key', transfers.getPublicKey)
```
MatthewPhinney: I put this endpoint under transfers to emphasize that it returns the ed25519 key associated with the transfer condition.

src/lib/app.js
emschwartz: Wait, you mean return the hash, not the hash preimage. The preimage is the "fulfillment" / what you get when it's all done

src/controllers/transfers.js
clark800: I think you should only make a generator if necessary

src/models/transfers.js
```diff
@@ -74,6 +74,24 @@ function * getTransferStateReceipt (id, signatureType, conditionState) {
...
+    public_key: config.getIn(['keys', 'ed25519', 'public'])
+  }
+}
+
+function * getPreimage (transferId, state) {
+  if (!state) {
+    const transfer = yield db.getTransfer(transferId)
+    state = transfer ? transfer.state : 'nonexistent'
+  }
+  const conditionPreImage = makePreImage(transferId, state)
```
clark800: Sometimes it's `PreImage` sometimes `Preimage`

src/lib/app.js
emschwartz: This seems odd to me. Are there multiple keys? Even if there are it would seem more sensible to have an endpoint for getting keys and then have them listed based on what they're for.

Also, is it dangerous to make anyone rely on getting the public key over HTTPS? What if the signature scheme the ledger is using is more secure than TLS?

sentientwaffle:
Is this supposed to replace `/transfers/:transfer_id/state`, or just supplement it?

src/lib/app.js
MatthewPhinney: Yes, it returns a hash. The preimage of the hash is
```
{
    id: transfer_id,
    state: state,
    token: sign(sha512(transfer_id + ':' + state))
  }
```
Can you clarify how the hash is used by the sender?

MatthewPhinney:
> Is this supposed to replace /transfers/:transfer_id/state, or just supplement it?

I was actually wondering that, as well.

src/lib/app.js
emschwartz: The hash is used as the condition. The preimage is the signature

MatthewPhinney: How does anyone aside from the ledger know the preimage? Is this implemented in `five-bells-condition`?

sentientwaffle: No one besides the ledger is supposed to know the preimage until the condition is met, at which point the ledger attaches it as the fulfillment.

sentientwaffle:
@emschwartz: we need to distribute the ledgers' public key somehow (and the notaries' as well) -- is there a better way than over HTTPS?

I talked to @justmoon and I'm going to add the endpoint `/` that returns metadata about the ledger/notary, including the public_key.

@MatthewPhinney: `/transfers/:id/hash_preimage` is a misleading name, since it returns the digest and not the preimage. `getPreimage` returns a state+digest, not the preimage.

----

Instead of splitting `/transfers/:id/hash_preimage` out of `/state`, what about:

* `/transfers/:id/ed25519-sha512`: like /state, but no `type` parameter. Returns `{type, signer, public_key, message: {id,state}, signature}`
* `/transfers/:id/sha256`:
  * If `condition_state` is passed, returns `{type, signer, digest}` (where digest is the hash of the preimage that includes that state).
  * If `condition_state` is _not_ passed, returns `{type, signer, digest, message}` (where message is the preimage with the current state, and digest is the hash of that preimage).
* Remove the `/state` endpoint entirely.

This would also simplify the signed message schema.

justmoon:
> `/transfers/:id/ed25519-sha512`

What is that? Based on the URL I would guess it's the transfer object signed with ed25519. I definitely wouldn't expect it to be a different transfer state resource.

The crypto-conditions work is likely going to lead to a total overhaul of the /state endpoint. So here is what I suggest:

1. Add the `GET /` endpoint which returns a ledger resource that contains name of the ledger, currency name, currency symbol, currency code (all optional, configurable fields), public key (addressing RILP-95) and links to other API endpoints (compare https://api.github.com/)
2. Don't touch the /state endpoint for now. No need to make breaking changes to something that is about to be totally overhauled.

justmoon:
I'm closing this PR because the decision was not to change this right now pending deployment of the new crypto-conditions.

PR #168, opened 2016-02-24
[FEATURE] metadata endpoint: /

sentientwaffle:
Return metadata from `/`, which includes the ledger's `public_key` among other things.

/cc @justmoon

### DIFF
--- a/src/controllers/metadata.js
+++ b/src/controllers/metadata.js
@@ -1,0 +1,31 @@
+'use strict'
+const config = require('../services/config')
+
+const metadata = {
+  currency_code: config.getIn(['currency', 'code']),
+  currency_symbol: config.getIn(['currency', 'symbol']),
+  public_key: config.getIn(['keys', 'ed25519', 'public']),
+  urls: {
+    health: '/health',
+    transfer: '/transfers/:id',
+    transfer_fulfillment: '/transfers/:id/fulfillment',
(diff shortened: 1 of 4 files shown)


src/controllers/metadata.js
```diff
@@ -0,0 +1,34 @@
...
+  health_url: '/health',
+  transfer_url: '/transfers/:id',
+  transfer_fulfillment_url: '/transfers/:id/fulfillment',
+  transfer_state_url: '/transfers/:id/state',
+  connectors_url: '/connectors',
+  accounts_url: '/accounts',
+  account_url: '/accounts/:name',
+  subscription_url: '/subscriptions/:id',
+  subscription_notification_url: '/subscriptions/:subscription_id/notifications/:notification_id'
+}, function (path) { return baseURI + path })
```
clark800: Does it have to be the full URL? Relative URLs tend to be simpler because they don't have issues with proxies etc. Also, do you think it would be better to keep the URLs in their own sub-object?

sentientwaffle: Sure, I was just using https://api.github.com/ as an example.

sentientwaffle: Ok, added a sub-object for the URLs, and switched to relative paths.

clark800:
LGTM

PR #172, opened 2016-02-25
TEST add debug logging

MatthewPhinney:
Do not merge -- this is for debugging purposes only.


MatthewPhinney:
Closing this, as it looks like the timeouts are due to the migrations/db sync done in the before each hook. Since we can safely say that this will not be an issue in production, we can solve this by increasing the test timeouts. This is tracked in RILP-217.

PR #174, opened 2016-02-26
Default admin account created with TLS certificate fingerprint

lumberj:
To allow admin account to authenticate using client certificate


naoitoi:
Would we allow setting both password hash and certificate fingerprint? 
It seems to me the admin should use either use certificate **or** password.  If a customer configures fingerprint, I'd like to eliminate the possibility that the password is accidentally set and be used for authentication.

lumberj:
The client would set `LEDGER_AUTH_BASIC_ENABLED` to `false`

MatthewPhinney:
LGTM

naoitoi:
I had a detailed discussion with @lumberj, and we agreed this PR is fine.  We will make another modification in authentication code:

> Another approach may be - user can set both pwd and fingerprint, but if you have fingerprint set, you can only log in with cert.

LGTM.

Issue #173, opened 2016-02-26
Authorized credits


justmoon:
There is a common use case we've talked about which is: As a recipient I want to approve all incoming transfers.

For example:

* As a bank I don't want transfers incoming on the ILP level that Ripple Connect doesn't expect.
* As an ILP-enabled wallet provider I don't want transfers incoming on the ILP level that don't have appropriate metadata for a higher-level protocol, such as sender information.

## Functional specification

To allow for this, we add a new field to credits called `authorized`. Like the `authorized` field on debits, it can only be set by the owner of the account that that credit references (or an admin.) If this feature is turned on, a transfer cannot progress to `prepared` unless all debits and credits are authorized.

The feature is disabled by default (to avoid a breaking change) and can be enabled globally with an environment variable `LEDGER_FEATURE_CREDIT_AUTH=true`.

PR #176, opened 2016-02-29
[FEATURE] credit auth

sentientwaffle:
Fixes https://github.com/interledger/five-bells-ledger/issues/173

It is disabled by default, so this is not a breaking change.

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -11,7 +11,12 @@ function isRunningTests () {
 }
 
 function loadConfig () {
-  const localConfig = {}
+  const localConfig = {
+    features: {
+      hasCreditAuth: Config.getEnv('FEATURE_CREDIT_AUTH')
+    }
+  }
+
   const admin_user = Config.getEnv(envPrefix, 'ADMIN_USER') || 'admin'
   const admin_pass = Config.getEnv(envPrefix, 'ADMIN_PASS')
   const admin_fingerprint = Config.getEnv(envPrefix, 'ADMIN_TLS_FINGERPRINT')
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -13,7 +13,7 @@ function isRunningTests () {
 function loadConfig () {
   const localConfig = {
     features: {
-      hasCreditAuth: Config.getEnv('FEATURE_CREDIT_AUTH')
+      hasCreditAuth: Config.castBool(Config.getEnv('FEATURE_CREDIT_AUTH'))
     }
   }
 
--- a/src/models/transfers.js
+++ b/src/models/transfers.js
@@ -184,19 +184,23 @@ function validateIsAffectedAccount (account, transfer) {
   }
 }
 
-function validateAuthorizations (authorizedAccount, transfer, previousDebits) {
-  if (previousDebits && transfer.debits.length !== previousDebits.length) {
-    throw new UnprocessableEntityError('Invalid change in number of debits')
+/**
+ * @param {Account} authorizedAccount
+ * @param {Funds} funds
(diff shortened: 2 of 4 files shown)


src/models/transfers.js
```diff
@@ -181,19 +181,23 @@ function validateIsAffectedAccount (account, transfer) {
...
-function validateAuthorizations (authorizedAccount, transfer, previousDebits) {
-  if (previousDebits && transfer.debits.length !== previousDebits.length) {
-    throw new UnprocessableEntityError('Invalid change in number of debits')
+/**
+ * @param {Account} authorizedAccount
+ * @param {Funds} funds
+ * @param {Funds|undefined} previousFunds
+ * @param {String} type "debit" or "credit"
+ */
+function validateAuthorizations (authorizedAccount, fundsList, previousFunds, fundsType) {
```
clark800: Another naming idea for this: we used `adjustment` as a hypernym of `debit` and `credit` in RippleAPI

MatthewPhinney:
Code LGTM, although in general I'm kind of wary of introducing command line flags that change the behavior of the app. Maybe one flag is fine, but as we add more, the app gets much more difficult to debug. If we have even two or three of these flags, we could end up with the exact same code producing quite different results, depending on what flags are enabled. How strong is the business use case for allowing this to be enabled/disabled? Is this something that we definitely don't want in all instances of the app?

src/lib/config.js
```diff
@@ -10,7 +10,12 @@ function isRunningTests () {
 }
 
 function loadConfig () {
-  const localConfig = {}
+  const localConfig = {
+    features: {
+      hasCreditAuth: Config.getEnv('FEATURE_CREDIT_AUTH')
```
lumberj: `Config.castBool(Config.getEnv(envPrefix, 'FEATURE_CREDIT_AUTH'))`

src/models/transfers.js
sentientwaffle: I'm open to that, but its out of the scope of this PR -- the name 'funds' refers to the schema in five-bells-shared.

.eslintrc
```diff
@@ -1,2 +1,4 @@
 # We use JavaScript Standard Style (https://github.com/feross/standard)
 extends: standard
+rules:
```
clark800: Unnecessary now

clark800:
@sentientwaffle This pull request has been sitting idle for a while, can you respond to the questions?

sentientwaffle:
@MatthewPhinney https://github.com/interledger/five-bells-ledger/issues/173 specifies an env variable feature flag, in order to prevent a breaking change.

@lumberj good catch; fixed

test/putTransferSpec.js
```diff
@@ -1263,4 +1264,84 @@ describe('PUT /transfers/:id', function () {
       .expect(401)
       .end()
   })
+
+  describe('FEATURE_CREDIT_AUTH=true', function () {
+    // Kludge to work around immutable config
```
lumberj: This is no longer necessary with five-bells-shared@12

lumberj: (config mutable during unit testing)

justmoon:
@MatthewPhinney Good comment. I think in this situation a feature flag is the right approach. Eventually we probably want this to be always enabled, but there are a lot of changes that need to happen to other components. So adding this behind a feature flag for now allows us to start working on those other components without any time pressure from an open PR or a broken implementation.

When everything has migrated to the new behavior we can start defaulting to having this feature enabled and eventually remove the flag altogether.

MatthewPhinney:
Makes sense. LGTM.

PR #177, opened 2016-02-29
Replace sequelize with knex

clark800:


src/lib/app.js
```diff
@@ -56,8 +57,9 @@ class App {
     yield this.timerWorker.start()
     yield this.notificationWorker.start()
 
-    if (this.config.getIn(['db', 'sync'])) yield this.db.sync()
-    yield this.db.init()
+    if (this.config.getIn(['db', 'sync'])) {
```
naoitoi: Notary uses environment variable process.env.NOTARY_RUN_MIGRATION to initiate migration.  
Perhaps we should change either Notary or Ledger to be consistent with how to initiate migration.

src/lib/notificationWorker.js
```diff
@@ -31,25 +43,10 @@ class NotificationWorker {
...
-          }]
-        }, {
-          subject: {
-            $in: affectedAccounts
-          }
-        }]
-      },
-      transaction
-    })
+    let subscriptions = yield transaction.from('subscriptions')
```
naoitoi: I had an issue with using yield with transaction on sqlite, so I didn't use any transaction for sqlite on Notary.
If this works with sqlite for you, that's fine, but please keep an eye on it.

src/migrations/201602231400_initial.js
```diff
@@ -0,0 +1,107 @@
+'use strict'
+
+function createAccountsTable (knex) {
```
naoitoi: As discussed with @clark800 , I would recommend some schema changes, such as not using enum, preferring varchar to text, normalizing what's in JSON column.  But we agreed that we'd do it in another ticket, which I will file.

src/lib/notificationWorker.js
clark800: Using transactions is a requirement in the ledger, so if there is a problem we will have to figure it out

src/migrations/201602231400_initial.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function createEntriesTable (knex) {
+  return knex.schema.createTableIfNotExists('entries', table => {
+    table.increments()
+    table.integer('entry_group')
+    table.uuid('transfer_id')
+    table.integer('account')
+    table.decimal('balance', 10, 2)
+    table.datetime('created_at')
```
naoitoi: I like using table.timestamps() and letting Knex create created_at and update_at, rather than creating them manually.  But I'm OK with this implementation.

src/models/db/account.js
```diff
@@ -69,6 +67,8 @@ class Account extends Model {
   }
 
   static convertFromPersistent (data) {
+    data.is_disabled = Boolean(data.is_disabled)
```
naoitoi: We may want to explain the conversion necessary for booleans in a comment?

src/models/db/conditionFulfillment.js
```diff
@@ -23,14 +20,21 @@ class ConditionFulfillment extends Model {
   static convertFromPersistent (data) {
     delete data.created_at
     delete data.updated_at
+    if (typeof data.condition_fulfillment === 'string') {
```
naoitoi: Perhaps adding a line of comment to explain why we'd like to selectively convert a string to JSON might help?

src/models/transfers.js
```diff
@@ -286,7 +286,7 @@ function * fulfillTransfer (transferId, fulfillment) {
     const isExecution = validateConditionFulfillment(transfer, fulfillment)
 
     if (isExecution && transfer.state === 'executed' || !isExecution && transfer.state === 'rejected') {
-      return (yield fulfillments.getFulfillment(transferId)).getDataExternal()
+      return (yield fulfillments.getFulfillment(transferId, {transaction})).getDataExternal()
```
naoitoi: Why would you like to take a transaction?  If getFulfillment() simply does select, transaction doesn't seem to be necessary?

src/lib/notificationWorker.js
naoitoi: Knex has two ways of using transaction.  http://knexjs.org/#Transactions

1.    As the "promise aware" knex connection
 2.  As an object passed into a query with and eventually call commit or rollback.

Method (2) didn't work for me with a generator (see: https://github.com/tgriesser/knex/issues/811), so I exclusively used (1) in Knex.  You are using both methods ... Has (2) been tested thoroughly (e.g. take a transaction, do some insert or update operation, throw an exception and see if the transaction is rolled back)?

naoitoi:
Thank you for creating this PR.  I can see a lot of work went into this.

src/migrations/201602231400_initial.js
MatthewPhinney: I like that approach. I think we can use `table.timestamps() ` everywhere.

src/migrations/201602231400_initial.js
```diff
@@ -0,0 +1,107 @@
...
+    table.integer('retry_count')
+    table.datetime('retry_at').index()
+    table.index(['subscription_id', 'transfer_id'], 'subscription_transfer')
+  })
+}
+
+function createSubscriptionsTable (knex) {
+  return knex.schema.createTableIfNotExists('subscriptions', table => {
+    table.uuid('id').primary()
+    table.string('owner', 1024)
```
MatthewPhinney: Why use 1024 as the length? I noticed, in other tables, some `string` datatypes don't have a length specified.

src/models/db/account.js
MatthewPhinney: A comment would be helpful. Also, why convert to a number below (in `convertToPersistent`)?

src/models/db/subscriptions.js
```diff
@@ -9,12 +9,11 @@ function * getSubscription (id, options) {
   return (yield Subscription.findById(id, options))
 }
 
-function * getMatchingSubscription (subscription, options) {
-  return (yield Subscription.findOne({where: {
+function * getMatchingSubscriptions (subscription, options) {
+  return (yield Subscription.findWhere({
```
MatthewPhinney: This should, in principle, return only 1 subscription. See here: https://github.com/interledger/five-bells-ledger/issues/70; and here: https://github.com/interledger/five-bells-ledger/pull/85.

src/models/db/transfer.js
```diff
@@ -72,11 +69,44 @@ class Transfer extends Model {
...
+      data.execution_condition = JSON.parse(data.execution_condition)
+    }
+    if (typeof data.cancellation_condition === 'string') {
+      data.cancellation_condition = JSON.parse(data.cancellation_condition)
+    }
+    if (typeof data.additional_info === 'string') {
+      data.additional_info = JSON.parse(data.additional_info)
+    }
+    if (data.expires_at) {
+      data.expires_at = new Date(data.expires_at)
```
MatthewPhinney: Why do we need to call the `Date` constructor here, but not in `convertToPersistent`?

src/models/db/transfers.js
```diff
@@ -7,8 +9,32 @@ function * getTransfer (id, options) {
...
+  } else {
+    yield Transfer.create(transfer, options)
+  }
+  return Boolean(existingTransfer)
+}
+
 function * upsertTransfer (transfer, options) {
-  yield Transfer.upsert(transfer, options)
+  if (options && options.transaction) {
+    return (yield _upsertTransfer(transfer, options))
```
MatthewPhinney: In some cases, it is `return yield <expression>` and in others `return (yield <expression>)`.

src/models/db/transfers.js
```diff
@@ -7,8 +9,32 @@ function * getTransfer (id, options) {
   return (yield Transfer.findById(id, options))
 }
 
+function * _upsertTransfer (transfer, options) {
+  assert(options.transaction)
+  // SQLite's implementation of upsert does not tell you whether it created the
```
MatthewPhinney: I think this comment is a little dated. Since we are not actually calling `upsert` on a db model, maybe just say that we need the correct status code (200/201), so we need to know if the transfer exists in the db.

src/models/notifications.js
```diff
@@ -7,27 +7,29 @@ const UnauthorizedError = require('five-bells-shared/errors/unauthorized-error')
...
 const subscriptionUtils = require('../lib/subscriptionUtils')
 
 function * getNotification (subscriptionId, notificationId, requestingUser) {
   log.debug('fetching notification ID ' + notificationId)
-  const notification = yield Notification.findOne({
-    where: {
-      id: notificationId
-    },
-    include: [{ all: true }]})
+  const notification = yield Notification.findById(notificationId)
```
MatthewPhinney: We may be able to do this more easily with a join (so we don't need two more queries to get the subscription and transfer), but I think it's fine to leave as is, for this PR.

src/models/db/transfer.js
naoitoi: new Date() is needed before writing to Oracle DB.  So - I agree with Matt, it seems new Date() would be needed more in convertToPersistent() than in convertFromPersistent().
But I think we'll find out for sure once we exercise this code in test with Oracle.

src/models/subscriptions.js
```diff
@@ -31,9 +31,9 @@ function * setSubscription (subscription, requestingUser) {
 
   let existed
   yield db.transaction(function * (transaction) {
-    const duplicate = yield db.getMatchingSubscription(
+    const duplicates = yield db.getMatchingSubscriptions(
```
MatthewPhinney: Since we throw an error when `getMatchingSubscriptions` returns any record (so we never expect it to return more than 1), I'd probably change the name back to `getMatchingSubscription`. This logic could be cleaned up in general, but that's outside the scope of this PR.

test/accountSpec.js
```diff
@@ -298,47 +299,4 @@ describe('Accounts', function () {
       expect(user.public_key).to.equal(publicKey)
     })
   })
-
-  describe('Account#findEntry', function () {
```
MatthewPhinney: Why don't we need this anymore?

src/models/transfers.js
clark800: This inside a transaction block, everything inside the block should be in the same transaction.

src/migrations/201602231400_initial.js
clark800: Yeah, seems like a good idea. I'll have to check if we have to add a `defaultsTo` clause (from the docs: `  table.timestamp('created_at').defaultTo(knex.fn.now());`)

src/migrations/201602231400_initial.js
clark800: This is an attempt to keep it as close as possible to the previous schemas. The default for the `string` type is 255.

src/models/db/account.js
clark800: Yeah, I'll add a comment. Some databases store booleans as `0` and `1` so it has to be converted back. I probably shouldn't convert to number in `convertToPersistent`, I'll try changing that.

src/models/db/subscriptions.js
clark800: Yes, the name is because it returns an array type of length 0 or 1. I'll refactor to make it more clear.

src/models/db/transfer.js
clark800: I think because the standard format is a date object. So when you get it from the database you have to parse to a date, but when putting it into the database it automatically converts the date to the proper format.

test/accountSpec.js
clark800: It was never used

src/lib/app.js
clark800: Yeah, good idea, though out of scope for this pull request

src/migrations/201602231400_initial.js
clark800: On second thought, I think being explicit is better; it looks like `timestamps` is just a shorthand and I don't think it improves anything.

src/models/db/conditionFulfillment.js
clark800: I'm going to switch to text fields and remove the conditionals

src/lib/notificationWorker.js
clark800: I tested that transactions are working manually and had Nao review the method that I was using to test.

naoitoi:
LGTM.  

Thank you very much for the hard work, @clark800 !

PR #178, opened 2016-03-01
Use system dictionary for transfer states

MatthewPhinney:


naoitoi:
How about defining the states in -shared, so that all the ILP components can use the same dictionary?
Inter-component consistency ...

MatthewPhinney:
I'm open to that idea. You could argue that we already do define the states in shared: https://github.com/interledger/five-bells-shared/blob/master/schemas/TransferState.json. We could move the transferState file into shared/lib.

MatthewPhinney:
@naoitoi updated to use the library in five-bells-shared@11.1.0.

src/lib/transferExpiryMonitor.js
```diff
@@ -65,9 +68,8 @@ class TransferExpiryMonitor {
         log.debug('transfer ' + transfer.id +
           ' will expire in ' + expiry.diff(now, 'milliseconds') + 'ms')
       }
-    } else if (transfer.state === 'executed' ||
-      transfer.state === 'rejected' ||
-      transfer.state === 'failed') {
+    } else if (transfer.state === transferStates.TRANSFER_STATE_EXECUTED ||
+      transfer.state === transferStates.TRANSFER_STATE_REJECTED) {
```
naoitoi: Is it OK not to check for 'failed'?

naoitoi:
LGTM other than:
- Changing the way how we require(), as discussed offline, and
- Failing unit test

src/lib/transferExpiryMonitor.js
MatthewPhinney: `failed` is no longer a valid transfer state, as per the schema: https://github.com/interledger/five-bells-shared/blob/master/schemas/TransferState.json.

MatthewPhinney:
I updated the requires to be consistent with `five-bells-shared@11.1.1`: `const transferDictionary = require('five-bells-shared').TransferStateDictionary`

naoitoi:
LGTM, thanks!

PR #182, opened 2016-03-02
Enable client-cert auth

lumberj:
Use authentication for all PUT /transfers
Disallow anonymous PUT /transfers on ledger


lumberj:
Splitting this up

PR #186, opened 2016-03-04
[TASK] remove EntryGroup

sentientwaffle:


src/migrations/201602231400_initial.js
```diff
@@ -22,17 +22,10 @@ function createFulfillmentsTable (knex) {
...
-    table.increments()
-    table.timestamp('created_at').defaultTo(knex.fn.now())
-  })
-}
-
 function createEntriesTable (knex) {
   return knex.schema.createTableIfNotExists('entries', table => {
     table.increments()
-    table.integer('entry_group')
+    table.string('group', 1024)
```
clark800: Isn't `transfer_id` enough?

sentientwaffle: Nope, a single transfer can have 2 entry groups: one to apply debits, and another that either applies credits or reverts the debits.

clark800: Why do we need them grouped that way?

sentientwaffle: It allows you to get the balances of all accounts at any point in time and ensure that they sum to a constant value.

If they were only collected by `created_at` and not `group`, it would include extra entries, so it wouldn't sum to a constant.

If they were collected by `transfer_id` you wouldn't be able to perform the operation at any point in time, only after any transfer completes.

clark800: > If they were collected by transfer_id you wouldn't be able to perform the operation at any point in time, only after any transfer completes.

Are we not recording entries for the hold account?

clark800:
LGTM

PR #188, opened 2016-03-07
[TASK] Use NotificationScheduler

sentientwaffle:


.eslintrc
```diff
@@ -1,2 +1,4 @@
 # We use JavaScript Standard Style (https://github.com/feross/standard)
 extends: standard
+rules:
```
clark800: This looks like a slippery slope

sentientwaffle: I had to add that to get the linter to pass.

clark800: So fixing the errors was not possible?

sentientwaffle: This PR is for `NotificationScheduler`, not for fixing 143 eslint errors spread across the entire project.

clark800: You can use the ratchet mechanism that we used in ripple-lib so that you only have to fix errors in files that you modify, which makes these situations a lot less painful: https://github.com/ripple/ripple-lib/blob/develop/scripts/ci.sh#L22

clark800: Looks like Alan already took care of it though: https://github.com/interledger/five-bells-ledger/pull/190/files

clark800:
LGTM

PR #190, opened 2016-03-08
Fix lint errors

lumberj:
Spacing after * in generator functions


MatthewPhinney:
This is because of the new eslint version? I think I was out when we put that change in.

lumberj:
Yes...it's started failing for me locally

lumberj:
>  I'm not tracking why this wasn't caught by the npm run lint task in circleci.

I'm confused by that as well. devDependencies are not strictly versioned and we depend on a plugin that sets the lint rules, so that can change under us.

PR #193, opened 2016-03-08
Allow negative account balances (RILP-263)

MatthewPhinney:


src/lib/accountBalances.js
```diff
@@ -38,7 +38,8 @@ AccountBalances.prototype._getAccountBalances = function * (creditsOrDebits) {
 
     accounts[account] = {
       balance: +accountObj.balance,
-      totalAmount: +_.sum(_.map(amounts, parseFloat))
+      totalAmount: +_.sum(_.map(amounts, parseFloat)),
```
clark800: I don't think the `+` sign is necessary

MatthewPhinney: I think it was left over from before we moved to knex. I'll remove it.

clark800:
LGTM

PR #194, opened 2016-03-11
Fixes to support Oracle

clark800:


src/lib/knex.js
```diff
@@ -1,25 +1,17 @@
...
-    if (process.env.LEDGER_UNIT_DB_URI.startsWith('mysql:')) {
-      return 'mysql'
-    }
-    if (process.env.LEDGER_UNIT_DB_URI.startsWith('postgres:')) {
-      return 'postgres'
-    }
-  }
-  return 'development'
-}
+const config = require('../services/config')
```
naoitoi: Is it OK for delete this logic, i.e., parse URL and set knex environment?

package.json
```diff
@@ -20,7 +20,8 @@
     "coveralls": "cat ./coverage/lcov.info | coveralls",
     "apidoc": "apidoc -o apidoc-out -i src/controllers/",
     "integration": "integration all",
-    "bump": "version-bump"
+    "bump": "version-bump",
+    "drop-oracle-tables": "scripts/drop-oracle-tables.sh"
```
MatthewPhinney: So the reason for this script is that you can't pass env variables to to `knex`? Is this why `DYLD_LIBRARY_PATH=/opt/oracle/instantclient knex --knexfile=src/knexfile.js migrate:latest` doesn't work?

scripts/drop-oracle-tables.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export DYLD_LIBRARY_PATH=/opt/oracle/instantclient/
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+/opt/oracle/instantclient/sqlplus system/oracle@192.168.99.100:49161/xe < "$DIR/drop-tables.sql"
```
naoitoi: Minor comment:  I would take username and password from the command line arguments, so that you can drop table as a different Oracle user.  But if the purpose of this script is primarily to drop tables as "system" user, that's fine.

package.json
clark800: You mean as opposed to inlining it? I made a separate file because it's easier to read that way.

src/lib/knex.js
```diff
@@ -1,25 +1,17 @@
...
 
 // This module encapsulates Knex instance, which is used to access database.
 // Environment variable LEDGER_DB_ENV should be set according to deployment type
 // E.g., 'production' or 'staging'.
 // This is used to look up Knex configuration in knexfile.js
-const knexConfig = require('../knexfile')[getKnexConfigType()]
+const knexConfigEnv = config.getIn(['db', 'knex_env'])
```
MatthewPhinney: Is LEDGER_DB_ENV a new env variable? If so, can you add it to the readme?

scripts/drop-oracle-tables.sh
clark800: This repo only uses the "system" user, so this should be sufficient.

src/lib/knex.js
clark800: Nao decided to rename LEDGER_DB_ENV to LEDGER_DB_KNEX_ENV. I'll update the comment and readme.

src/lib/knex.js
clark800: I think this isn't being done correctly for mysql and postgres. I'll try to think of a good solution.

src/models/db/account.js
```diff
@@ -83,7 +83,9 @@ class Account extends Model {
     // some databases store booleans as 0 and 1, and knex does not convert
     data.is_disabled = Boolean(data.is_disabled)
     data.is_admin = Boolean(data.is_admin)
-    data.balance = Number(data.balance)
+    // oracle has balance stored properly, but knex returns it with a small
+    // rounding error, possibly a bug in knex; using toFixed as workaround
+    data.balance = Number(data.balance.toFixed(2))
```
naoitoi: toFixed(2) implies @yana-novikova

naoitoi: toFixed(2) means we will always round to 0.01 of the currency (i.e. to the cent if the currency is USD).  Is this acceptable @yana-novikova ?  Also, if we round balance to toFixed(2) here, is there an implication on other components?  For example, if Connector says "please add this account $1.005" to Ledger, and Ledger adds $1.01, would there be a problem for Connector?

clark800: I think the database was already enforcing this constraint. Though it looks like the integration tests are failing due to this.

scripts/drop-oracle-tables.sh
lumberj: This is hardcoding the IP address for Oracle. On my system, Oracle server is not running at the URL

scripts/drop-oracle-tables.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
```
lumberj: Use `set -ex pipefail` so it exits early on any subprocess that has non-zero exit code

scripts/drop-oracle-tables.sh
MatthewPhinney: I agree this is bad practice, but we already have hardcoded IPs in the knexfile. We should probably change all of them in one go.

src/knexfile.js
```diff
@@ -73,7 +73,7 @@ module.exports = {
   },
   // Test with Oracle on Linux (e.g. on CircleCI)
   // Set environment variables
-  // NOTARY_DB_ENV=oracledci LD_LIBRARY_PATH='/opt/oracle/instantclient'
+  // LEDGER_DB_KNEX_ENV=oracledci LD_LIBRARY_PATH=/opt/oracle/instantclient
```
lumberj: What about not hardcoding the location of instantlclient either? Or at least setting this as default and letting me use an ENV variable to change it.

scripts/drop-oracle-tables.sh
clark800: There's only one command that can fail here

src/knexfile.js
clark800: We should probably require that `sqlplus` is in `$PATH` and create a standardized setup process, but I think that can be a separate task.

lumberj: Proper instructions to setup Oracle in the README would help too. Agreed it can be a separate task.

clark800:
Closing in order to create pull request from interledger repo in order to make integration test branch coordination work

PR #195, opened 2016-03-14
FIX: Floating point precision errors when applying debits/credits

lumberj:
Uses bignumber.js

```js
> 94.9899 - 5.0101
89.97980000000001
```

vs.

```js
> (new Bignumber(94.9899)).minus(5.0101).toNumber()
89.9798
```

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -19,6 +19,11 @@
         }
       }
     },
+    "bignumber.js": {
+      "version": "2.3.0",
+      "from": "bignumber.js@latest",
+      "resolved": "https://registry.npmjs.org/bignumber.js/-/bignumber.js-2.3.0.tgz"
+    },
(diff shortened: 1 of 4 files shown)


lumberj:
@clark800 @MatthewPhinney

MatthewPhinney:
Nice use of the lodash api. To be clear, `sum()` and `difference()` return `Bignumber`, which is then written to the db as a js `Number`?

lumberj:
> Nice use of the lodash api. To be clear, sum() and difference() return Bignumber, which is then written to the db as a js Number?

No they written a Number. This is still a problem if we have numbers with more than 15 significant digits though...

lumberj:
Also, the tests don't pass because the CI database is using 2 decimal point precision, and sqlite doesn't enforce any. We are going to need to think about rounding rules. Closing this so we can discuss that first.

PR #196, opened 2016-03-14
Fixes to support Oracle

clark800:


MatthewPhinney:
LGTM. Although, I think we should have a separate task to remove the hardcoded IP/port for the oracle db.

PR #197, opened 2016-03-14
FIX: Floating point precision errors when applying debits/credits

lumberj:
Uses bignumber.js

```js
> 94.9899 - 5.0101
89.97980000000001
```

vs.

```js
> (new Bignumber(94.9899)).minus(5.0101).toNumber()
89.9798
```

Keep numbers as strings

### DIFF
--- a/src/lib/accountBalances.js
+++ b/src/lib/accountBalances.js
@@ -5,6 +5,7 @@ const UnprocessableEntityError = require('five-bells-shared/errors/unprocessable
 const InsufficientFundsError = require('../errors/insufficient-funds-error')
 const log = require('../services/log')('account balances')
 const Account = require('../models/db/account').Account
+const Bignumber = require('bignumber.js')
 
 function AccountBalances (transaction, transfer) {
   this.transaction = transaction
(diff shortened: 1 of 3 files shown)


lumberj:
cc @justmoon

lumberj:
@MatthewPhinney

src/lib/accountBalances.js
```diff
@@ -22,6 +23,20 @@ AccountBalances.prototype.applyDebits = function * () { yield this._applyDebits(
 AccountBalances.prototype.applyCredits = function * () { yield this._applyCredits(this._credits) }
 AccountBalances.prototype.revertDebits = function * () { yield this._applyCredits(this._debits) }
 
+function sum () {
```
clark800: Why doesn't this just take an array?

lumberj: Want me to change it?

justmoon: Yeah, an array would be preferred.

justmoon:
Note that doesn't fully address the subject of ledger precision. We also need to think about making precision configurable, add explicit rounding and figure out what to use as the database schema (so that the database doesn't do any unwanted rounding.)

But this is an incremental change in the right direction. LGTM.

MatthewPhinney:
LGTM. Although, which issue is this for? RILP-239? I think this addresses that task, but, as Stefan said, not the full issue of ledger precision. Are we tracking the ledger precision issue separately?

lumberj:
> Note that doesn't fully address the subject of ledger precision. We also need to think about making precision configurable, add explicit rounding and figure out what to use as the database schema (so that the database doesn't do any unwanted rounding.)

I have a followup task for this

lumberj:
> Are we tracking the ledger precision issue separately?

What issue?

src/lib/accountBalances.js
lumberj: Fixed

src/lib/accountBalances.js
```diff
@@ -22,6 +23,20 @@ AccountBalances.prototype.applyDebits = function * () { yield this._applyDebits(
 AccountBalances.prototype.applyCredits = function * () { yield this._applyCredits(this._credits) }
 AccountBalances.prototype.revertDebits = function * () { yield this._applyCredits(this._debits) }
 
+function sum (numbers) {
+  const numStrings = _.map(numbers, String)
+  return _.reduce(_.rest(numStrings), (result, num) => {
```
clark800: From the lodash docs: "If accumulator is not given the first element of collection is used as the initial value."

So you don't have to use `first` and `rest`, just pass `numStrings` as the first argument.

clark800: Oh, I guess you would have to do the BigNumber cast in the function then, so nevermind.

clark800:
LGTM

PR #203, opened 2016-03-16
FIX: Lowercase name and id when creating accounts

lumberj:
Bug:

```js
PUT http://localhost:3001/accounts/Skyla
{"name":"Skyla","balance":"50","is_admin":false}
```

Returns
```js
{"id":"InvalidBodyError","message":"Account name must match the one in
the URL (strictEqual, expected: \"skyla\", actual: \"Skyla\")"}
```

### DIFF
--- a/src/models/db/account.js
+++ b/src/models/db/account.js
@@ -37,6 +37,8 @@ class Account extends Model {
     if (data.id) {
       data.name = uri.parse(data.id, 'account').name.toLowerCase()
       delete data.id
+    } else {
+      data.name = data.name.toLowerCase()
     }
 
     if (data.balance) {
--- a/test/accountSpec.js
+++ b/test/accountSpec.js
@@ -288,19 +288,18 @@ describe('Accounts', function () {
   describe('PUT /accounts/:uuid', function () {
     it('should return 201', function * () {
       const account = this.exampleAccounts.bob
-      // Passwords are not returned
-      delete account.password
+      const withoutPassword = _.omit(account, 'password')
       yield this.request()
         .put(account.id)
         .auth('admin', 'admin')
         .send(account)
         .expect(201)
-        .expect(account)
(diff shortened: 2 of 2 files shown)


clark800:
So this was not a bug?

lumberj:
No, it was reported as a bug so I added a test case to check the behavior.
Do we explicitly test anywhere else that the ledger lowercases the name?

MatthewPhinney:
LGTM. I can't think of any other places, aside from account creation, that lowercase names, off the top of my head. I think all other places use UUIDs.

test/accountSpec.js
```diff
@@ -418,6 +417,42 @@ describe('Accounts', function () {
...
+        .expect(201)
+        .expect(withoutPassword)
+        .expect(validator.validateAccount)
+        .end()
+
+      // Check balances
+      expect((yield Account.findByName('bob')).getDataExternal()).to.deep.equal(withoutPassword)
+    })
+
+    it('should lowercase account name -- uppercased in body', function * () {
```
clark800: Please add a test case like this that omits the `id` field from the body; I think it will fail

lumberj: Good catch

lumberj: Updatd

src/models/db/account.js
```diff
@@ -37,6 +37,8 @@ class Account extends Model {
     if (data.id) {
       data.name = uri.parse(data.id, 'account').name.toLowerCase()
       delete data.id
+    } else {
+      data.name = data.name.toLowerCase()
```
clark800: Can you add a test case for PUTing an account that is missing both `id` and `name`? We need to make sure it isn't a 500 error.

lumberj: Yes

lumberj: Updated with test

PR #204, opened 2016-03-17
FEATURE Ledger enforces precision and scale on transfer amounts

lumberj:
Precision and scale allowed values exposed via metadata endpoint
Configured via env variables

Requires https://github.com/interledger/five-bells-integration-test/pull/13 and https://github.com/interledger/five-bells-connector/pull/113


lumberj:
This is failing the integration tests because the connector is setting precision after the decimal point to be `4` while the ledger by default will expect `2`:

https://github.com/interledger/five-bells-connector/blob/master/src/backends/fixerio/index.js#L143-L147

lumberj:
Requires: https://github.com/interledger/five-bells-connector/pull/112

justmoon:
It appears this is superceded by #215.

PR #205, opened 2016-03-17
[FIX] Seed: Hold account needs minimum_allowed_balance

vhpoet:


src/lib/seed-db.js
```diff
@@ -12,7 +12,7 @@ module.exports = function * (config) {
 function * setupHoldAccount () {
   const holdAccount = yield models.Account.findByName('hold')
   if (!holdAccount) {
-    yield models.Account.create({name: 'hold', balance: '0'})
+    yield models.Account.create({name: 'hold', minimum_allowed_balance: '0', balance: '0'})
```
clark800: The default value is 0 so this shouldn't have any effect, but it doesn't hurt to be explicit. If it was getting set to something besides 0 my current pull request will probably fix it.

clark800: I think both amounts are supposed to be numbers, but it looks like it will work either way.

clark800:
LGTM

PR #206, opened 2016-03-17
Fix overwriting of account settings when not specified in PUT

clark800:
Before this fix, if you PUT an account that just contains a password change it will zero the account's balance.

This was discovered while attempting to solve RILP-354, which was not reproducible.


gip:
LGTM

PR #212, opened 2016-03-18
Truncate tables between tests cases. Only migrate once

lumberj:
Should reduce unit test time significantly


lumberj:
cc @clark800

MatthewPhinney:
LGTM

PR #214, opened 2016-03-18
CI: Fix apidoc upload

lumberj:
The deploy task only runs on container(0) when builds are parallelized.
Artifacts generatd in other containers during test are not available.

CircleCI Support:

> Container 0 is the only one used for deployment, however prior to the
> deployment taking place all of the generated artifacts are already
> uploaded to S3 so if you need to use them you can pull them down or
> reference them from S3.


MatthewPhinney:
LGTM

PR #215, opened 2016-03-19
Feature: ledger precision is configurable and enforced on transfer amounts

lumberj:
Scale and precision configurable
`LEDGER_AMOUNT_SCALE`
`LEDGER_AMOUNT_PRECISION`

Throw exception if amount precision > 15 when using oracledb

Exposed via metadata endpoint

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -10,38 +10,47 @@ function isRunningTests () {
      process.argv[1].endsWith('mocha'))
 }
 
-function loadConfig () {
-  const localConfig = {
-    features: {
-      hasCreditAuth: Config.castBool(Config.getEnv(envPrefix, 'FEATURE_CREDIT_AUTH'))
-    }
-  }
+function useTestConfig () {
+  return !Config.castBool(process.env.UNIT_TEST_OVERRIDE) && isRunningTests()
+}
 
+function parseAmountConfig () {
+  return {
+    precision: parseInt(Config.getEnv(envPrefix, 'AMOUNT_PRECISION'), 10) || 10,
+    scale: parseInt(Config.getEnv(envPrefix, 'AMOUNT_SCALE'), 10) || 2
+  }
+}
+
+function parseCurrencyConfig () {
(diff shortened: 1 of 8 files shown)


lumberj:
Related integration test change: https://github.com/interledger/five-bells-integration-test/pull/14

MatthewPhinney:
LGTM (once all tests pass).

lumberj:
@MatthewPhinney there's a dependent integration test PR https://github.com/interledger/five-bells-integration-test/pull/14

src/lib/config.js
```diff
@@ -10,33 +10,47 @@ function isRunningTests () {
...
+}
+
+function parseAdminConfig () {
   const admin_user = Config.getEnv(envPrefix, 'ADMIN_USER') || 'admin'
   const admin_pass = Config.getEnv(envPrefix, 'ADMIN_PASS')
   const admin_fingerprint = Config.getEnv(envPrefix, 'ADMIN_TLS_FINGERPRINT')
 
   if (admin_pass || admin_fingerprint) {
-    localConfig.default_admin = _.omit({
+    return _.omit({
```
clark800: I think this is missing a return for the else case

lumberj: The `else` is that admin config is left undefined, same as original behavior

src/lib/config.js
```diff
@@ -44,8 +58,34 @@ function loadConfig () {
       }
     }
   }
+}
+
+function validateConfig () {
+  // Validate precision
+  const commonConfig = Config.loadConfig(envPrefix)
+  const isOracle = commonConfig.get('db.uri').match(/oracle/) !== null
```
clark800: Might be safer to test `startsWith('oracle://')`, the server name could contain "oracle".

src/lib/config.js
clark800: It would be more clear to have an explicit `return undefined`; otherwise the reader is wondering if it is intentional or not.

PR #216, opened 2016-03-20
Fix to bug RILP-356, RILP-388.  If notification received HTTP status …

naoitoi:
…422, do not resend.
I will discuss this with team before merging this.


justmoon:
-1

Let's be conservative. Even though any 4xx is technically a permanent failure, it could easily be caused by temporarily faulty components. We don't want to accidentally drop notifications.

The performance bug should be fixed in the connector. A valid notification hitting a correctly functioning connector should always result in a 2xx response. Whether or not the connector was waiting for that particular notification doesn't matter.

PR #221, opened 2016-03-22
[FEATURE | BREAKING] Ledger should sign notification (RILP-341)

MatthewPhinney:


lumberj:
@MatthewPhinney the integration test is failing

README.md
```diff
@@ -62,6 +62,8 @@ With the following configuration options set as environment variables:
 * `LEDGER_TLS_CERTIFICATE` (default: none) the path to the server certificate file. Required if using HTTPS.
 * `LEDGER_TLS_CRL` (default: none) the path to the server certificate revokation list file. Optional if using HTTPS.
 * `LEDGER_TLS_CA` (default: none) the path to a trusted certificate to be used in addition to using the [default list](https://github.com/nodejs/node/blob/v4.3.0/src/node_root_certs.h). Optional if using HTTPS.
+* `LEDGER_SIGNING_PRIVATE_KEY` the path to the file containing the private key used to sign ledger notifications.
```
lumberj: Add `(default: none)`

src/lib/config.js
```diff
@@ -65,12 +80,21 @@ function validateConfig () {
...
 
   // strong-oracle return native JS Numbers from Number type columns
   // Cannot support precision greater than 15
   if (!useTestConfig() && isOracle && amountConfig.precision > 15) {
     throw new Error('Cannot support precision > 15 with OracleDB')
   }
+
+  // Disallow the use of the same keys for TLS auth and notification signing
+  if (keys.notificationSign && keys.notificationSign.secret === tlsKey) {
+    if (keys.notificationSign.secret === tlsKey) {
```
lumberj: Is this a typo?

lumberj: What if they are both undefined?

test/configSpec.js
```diff
@@ -129,6 +135,14 @@ describe('loadConfig', () => {
         expect(_config.keys).to.deep.equal(testDefault)
       })
     })
+
+    it.skip('should throw an error when the TSL and ledger signing keys are the same', () => {
```
lumberj: Do you want to skip this test?

lumberj:
@MatthewPhinney are notification signing keys required? If so, this should be a breaking change (add `BREAKING` to commit title)

src/lib/config.js
lumberj: Also, right now tlsKey is a `Buffer` so I don't think using `===` will work

lumberj:
Right now the ledger generates an ED25519 keypair for signing crypto conditions if it's not running with `NODE_ENV=production`. (Presumably for the purposes of demoing).

Do we want to replicate this behavior for notification signing keypair?

MatthewPhinney:
> are notification signing keys required?

What do you think @naoitoi?

naoitoi:
> >  are notification signing keys required?
> What do you think @naoitoi?

I think notification signing is a requirement.  Without it, an attacker can trick Connector to falsely trust a untrustworthy notifications.

lumberj:
> I think notification signing is a requirement. Without it, an attacker can trick Connector to falsely trust a untrustworthy notifications.

Or, we can leave it up to the Connector

src/lib/config.js
```diff
@@ -55,26 +58,55 @@ function parseKeysConfig () {
         secret: 'iMx6i3D3acJPc4aJlK0iT/pkJP3T+Dqte9wg6hXpXEv08CpNQSm1J5AI6n/' +
           'QVBObeuQWdQVpgRQTAJzLLJJA/Q==',
         public: '9PAqTUEptSeQCOp/0FQTm3rkFnUFaYEUEwCcyyySQP0='
+      },
+      notificationSign: {
```
lumberj: I think the convention in config is to use snakecase

src/lib/config.js
```diff
@@ -55,26 +58,55 @@ function parseKeysConfig () {
...
+      }
+    }
+  } else {
+    const privateKeyPath = Config.getEnv(envPrefix, 'SIGNING_PRIVATE_KEY')
+    const publicKeyPath = Config.getEnv(envPrefix, 'SIGNING_PUBLIC_KEY')
+    if (!privateKeyPath) {
+      if (process.env.NODE_ENV === 'production') {
+        throw new Error('No ' + envPrefix.toUpperCase() + '_SIGNING_PRIVATE_KEY provided')
+      }
+      const keys = keypair()
```
lumberj: @clark800 you mentioned generating keys on headless could take a very long time. I haven't looked into where `keypair` gets its entropy, but that could be the case here, too. Maybe we should have a followup task for this and ed25519 keys in dev

lumberj:
Does verifying a JCS signature require knowledge of the public key? @MatthewPhinney

lumberj:
Let's adding the signing public key to the ledger's metadata endpoint. Similar to how we have included ed25519 key there.

https://github.com/interledger/five-bells-ledger/blob/master/src/controllers/metadata.js#L7

test/configSpec.js
```diff
@@ -129,6 +135,22 @@ describe('loadConfig', () => {
         expect(_config.keys).to.deep.equal(testDefault)
       })
     })
+
+    it('should throw an error when the TSL and ledger signing keys are the same', () => {
```
lumberj: typo

test/putTransferSpec.js
```diff
@@ -1016,13 +1016,13 @@ describe('PUT /transfers/:id', function () {
...
         const notificationId = idParts[idParts.length - 1]
-        expect(body).to.deep.equal({
+        expect(_.omit(body, 'signature')).to.deep.equal({
           event: 'transfer.update',
           id: subscription.id + '/notifications/' + notificationId,
           subscription: subscription.id,
           resource: transferResult
         })
-        // TODO validate this against notification schema
+        expect(validator.validateNotification.bind(validator.validateNotification, {body: body})).to.not.throw(Error)
```
lumberj: Why omit the signature? Shouldn't it be part of a valid Notification according to the schema (https://github.com/interledger/five-bells-shared/blob/master/schemas/Notification.json)

MatthewPhinney: If my understanding of `_.omit` is correct (it doesn't mutate the source object), it will omit the signature when checking the deep equality of the two objects (because the signature changes each time), but not when actually validating against the schema.

MatthewPhinney:
> Does verifying a JCS signature require knowledge of the public key?

@alan yes, see [here](https://github.com/interledger/five-bells-shared/blob/master/utils/jsonSigning.js#L110)

test/putTransferSpec.js
lumberj: You're right, I misread this

test/metadataSpec.js
```diff
@@ -25,6 +28,7 @@ describe('Metadata', function () {
             currency_code: null,
             currency_symbol: null,
             public_key: '9PAqTUEptSeQCOp/0FQTm3rkFnUFaYEUEwCcyyySQP0=',
```
lumberj: Since this is going to be a breaking change, I think maybe we should change this key to be condition_public_key or ed25519_public_key. It's otherwise confusing to have one generic and one that's specific

test/metadataSpec.js
```diff
@@ -17,6 +19,7 @@ describe('Metadata', function () {
 
   describe('GET /', function () {
     it('should return metadata', function * () {
+      const notificationPublicKey = fs.readFileSync(path.join(__dirname, './data/signKeyRSAPub.pem'), 'utf8')
```
lumberj: Why not use `config` here?

src/lib/config.js
```diff
@@ -55,26 +58,55 @@ function parseKeysConfig () {
...
+        notification_sign: {
+          secret: keys.private,
+          public: keys.public
+        }
+      }
+    }
+    return {
+      notification_sign: {
+        secret: fs.readFileSync(privateKeyPath, 'utf8'),
+        public: publicKeyPath ? fs.readFileSync(publicKeyPath, 'utf8') : undefined
```
lumberj: I don't think we should allow `public` to be undefined, since it's going to be shared in the metadata

src/lib/notificationWorker.js
```diff
@@ -115,17 +119,19 @@ class NotificationWorker {
         }
       }
     }
+    // Sign notification
+    const signedNotification = JSONSigning.sign(notificationBody, 'PS256', privateKey)
```
lumberj: Can you put PS256 into an explaining variable? 

```js
// RSA PSS signature with SHA-256
const signatureAlgo = 'PS256'
```

src/lib/config.js
MatthewPhinney: Good point.

MatthewPhinney:
@lumberj updated

src/lib/config.js
```diff
@@ -55,26 +59,57 @@ function parseKeysConfig () {
...
+    }
+  } else {
+    const privateKeyPath = Config.getEnv(envPrefix, 'SIGNING_PRIVATE_KEY')
+    const publicKeyPath = Config.getEnv(envPrefix, 'SIGNING_PUBLIC_KEY')
+    if (!privateKeyPath || !publicKeyPath) {
+      if (process.env.NODE_ENV === 'production') {
+        throw new Error(envPrefix.toUpperCase() + '_SIGNING_PRIVATE_KEY and ' +
+          envPrefix.toUpperCase() + '_SIGNING_PUBLIC_KEY must be provided')
+      }
+      log.info('Using autogenerated keys for notification signing.')
```
lumberj: Maybe put this at `WARN` level?

lumberj:
LGTM

src/lib/config.js
MatthewPhinney: haha ... I originally had this at warn level, then changed it to info. I'm fine with either. I'll update to `warn`, then merge the PR.

PR #222, opened 2016-03-22
Add unit test for creating account without password or fingerprint

clark800:
Covers case in RILP-349. It looks like the bug was already fixed.


MatthewPhinney:
LGTM

PR #223, opened 2016-03-22
Make PUT fulfillment anonymous only and GET fulfillment not anonymous

clark800:


naoitoi:
Could you provide some explanation?  What does this do, and why is it necessary?

clark800:
This is for RILP-287. There isn't a strong reason for the anonymous PUT other than making it impossible for the route handler to become user-dependent in the future. Making GET non-anonymous is to reduce the risk of denial of service attacks.

src/models/transfers.js
```diff
@@ -434,7 +434,10 @@ function * setTransfer (transfer, requestingUser) {
   }
 }
 
-function * getFulfillment (transferId) {
+function * getFulfillment (transferId, requestingUser) {
+  if (!requestingUser || !requestingUser.name) {
+    throw new UnauthorizedError('Not authorized')
```
MatthewPhinney: Do we want to check that the user is authorized to view the transfer that this fulfillment corresponds to?

clark800: I decided not to for now because it would require an extra database query and this is just to avoid DOS. It might be something we want in the future, I'm not sure.

MatthewPhinney:
LGTM

PR #224, opened 2016-03-23
Scalability fix.  On a typical dev environment (Oracle DB running on …

naoitoi:
…Docker on a laptop), keeping more than 2 Knex connections would cause problems.


MatthewPhinney:
LGTM

clark800:
Do we know why this causes problems? Is it a bug in knex? Is it a memory issue?

naoitoi:
Discussed with @clark800 offline - I think Oracle XE server, running on Docker, does not have enough resources (e.g. memory) to accommodate many concurrent connections.

PR #227, opened 2016-03-29
FEATURE Allow users to set some of their account properties

clark800:


emschwartz:
LGTM

MatthewPhinney:
Is this feature requirement documented in the spec? Where I should go for background info?

clark800:
I don't know, I found it here: https://ripplelabs.atlassian.net/browse/RILP-260

PR #229, opened 2016-03-30
Disable uploading apidocs for now

lumberj:
Will migrate to using Github pages
See:
https://github.com/interledger/five-bells-condition/blob/master/scripts/publish_spec.sh


justmoon:
LGTM

PR #231, opened 2016-04-11
fix: Allow setting db user and password from env variables

lumberj:
See:
https://github.com/interledger/five-bells-shared/blob/master/lib/config.js#L159-L161

### DIFF
--- a/src/lib/knex.js
+++ b/src/lib/knex.js
@@ -17,8 +17,8 @@ function parseKnexConnection (uri) {
   return {
     host: parsed.hostname,
     port: parsed.port,
-    user: auth[0],
-    password: auth[1],
+    user: auth[0] || config.get('db.connection_user'),
+    password: auth[1] || config.get('db.connection_password'),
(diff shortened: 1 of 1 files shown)


src/lib/knex.js
```diff
@@ -17,8 +17,8 @@ function parseKnexConnection (uri) {
   return {
     host: parsed.hostname,
     port: parsed.port,
-    user: auth[0],
-    password: auth[1],
+    user: auth[0] || config.get('db.connection_user'),
+    password: auth[1] || config.get('database.connection_password'),
```
MatthewPhinney: Why is one `db` and the other `database`?

lumberj: Typo

lumberj: Updated

MatthewPhinney:
LGTM

lumberj:
This method of authenticating to the DB is deprecated

PR #232, opened 2016-04-19
[PERF] Cache notification signatures

gip:


MatthewPhinney:
I don't think the notification body changes in any way between retries, to LGTM.

gip:
`notificationBody` is generated for each retry at https://github.com/gip/five-bells-ledger/blob/cache_signature/src/lib/notificationWorker.js#L106 @MatthewPhinney

emschwartz:
@gip If the notification body never changes, what about signing it when the notification is [originally created](https://github.com/interledger/five-bells-ledger/blob/15d9a867e2c0c07f1608531d21481ab9c006451f/src/lib/notificationWorker.js#L26) and storing the signature in the db instead? It seems a bit odd to me to have a special-purpose local cache for this.

PR #233, opened 2016-04-21
New crypto-conditions

justmoon:
This migrates Five Bells Ledger to the new crypto-conditions format. See https://github.com/interledger/five-bells-condition

Requires interledger/five-bells-shared#131


emschwartz:
LGTM (assuming tests are only failing because of shared dep)

Looks really good actually -- this is a very exciting change. Nice work on the conditions!

src/models/transfers.js
```diff
@@ -434,10 +447,7 @@ function * setTransfer (transfer, requestingUser) {
   }
 }
 
-function * getFulfillment (transferId, requestingUser) {
-  if (!requestingUser || !requestingUser.name) {
```
clark800: So `getFulfillment` allows anyone now? When does a non-registered user need to get a fulfillment?

clark800:
I think you have to update the version of five-bells-condition

src/lib/app.js
```diff
@@ -106,16 +106,8 @@ class App {
       models.Transfer.createBodyParser(),
       transfers.putResource)
 
-    router.put('/transfers/:id/fulfillment',
```
MatthewPhinney: what's the rationale for removing authentication?

src/models/transfers.js
justmoon: Since getting the transfer is unrestricted, getting the fulfillment should be as well for consistency. I think you can make a stronger argument for making the transfer details private than making the fulfillment private, since the fulfillment doesn't actually contain any meaningful information about the transfer.

So my thought was to remove the authentication from all transfer endpoints for now (simplicity/consistency) and then consider whether we want to add it to getting/putting transfers separately.

src/lib/app.js
justmoon: The PUT fulfillment call was allowing anonymous access anyway, so it doesn't make any sense to instantiate the passport middleware there. As for the GET fulfillment endpoint, I feel less strongly about the change, please see my response to @clark800 above for rationale.

src/models/transfers.js
clark800: Ok, I think we were planning on minimizing the anonymous endpoints to mitigate the risk of denial of service attacks, but this should be fine for now

clark800:
LGTM 👍

PR #234, opened 2016-04-27
Fix some new eslint issues

justmoon:
When clearing the cache on CircleCI, eslint-config-standard must have gotten updated and a new rule added. Because of that we have some test failures right now.


emschwartz:
LGTM

PR #235, opened 2016-04-27
Use crypto-conditions for notification signing

justmoon:
Using crypto-conditions speeds up notification signing and makes it more flexible (multi-sig, algorithm selection, etc.)


emschwartz:
LGTM (once tests pass)

src/controllers/metadata.js
```diff
@@ -1,21 +1,22 @@
...
-    health: '/health',
-    transfer: '/transfers/:id',
-    transfer_fulfillment: '/transfers/:id/fulfillment',
-    transfer_state: '/transfers/:id/state',
-    connectors: '/connectors',
-    accounts: '/accounts',
-    account: '/accounts/:name',
-    subscription: '/subscriptions/:id',
-    subscription_notification: '/subscriptions/:subscription_id/notifications/:notification_id'
+    health: base + '/health',
```
clark800: We intentionally left this out to prevent issues with proxies etc

lumberj: I think this is the publicly accessible URI for the server: https://github.com/interledger/five-bells-shared/blob/master/lib/config.js#L102

It comes from `CONNECTOR_HOSTNAME`, https://github.com/interledger/five-bells-shared/blob/master/lib/config.js#L86

lumberj: Just checked the README (https://github.com/interledger/five-bells-connector/#general)


> `CONNECTOR_HOSTNAME` (default: [your hostname]) Publicly visible hostname. This is important for things like generating globally unique IDs. Make sure this is a hostname that all your clients will be able to see. The default should be fine for local testing.

src/lib/notificationWorker.js
```diff
@@ -121,7 +121,7 @@ class NotificationWorker {
       }
     }
     // Sign notification
-    const algorithm = 'PS256' // RSASSA-PSS signature using SHA-256 and MGF1 padding with SHA-256
+    const algorithm = 'CC' // RSASSA-PSS signature using SHA-256 and MGF1 padding with SHA-256
```
lumberj: Is the comment still relevant?

src/controllers/metadata.js
clark800: Yes, but is there any advantage to using absolute URLs here? In general I think relative URLs are more robust because for example the same server can often be accessed from multiple hostnames (`localhost` and the DNS name is one common example). Even though it might work this way, the burden of proof should probably be on justifying using absolute URLs due to these considerations.

lumberj: Sorry, that was the Connector README, but still relevant:

https://github.com/interledger/five-bells-ledger#step-3-run-it

> LEDGER_HOSTNAME (default: [your hostname]) Publicly visible hostname. This is important for things like generating globally unique IDs. Make sure this is a hostname that all your clients will be able to see. The default should be fine for local testing.

justmoon: The ledger already has to know its public URI, because it uses it in transfer, account, etc. IDs. So base_uri is guaranteed to be the public URI after all proxies etc.

This is currently configured using the `LEDGER_PUBLIC_PATH`, `LEDGER_HOSTNAME`, `LEDGER_PUBLIC_PORT`, `LEDGER_PUBLIC_HTTPS` (from memory) config options. This grew over time, I'd say at some point we should probably just remove all these options and replace them with a `LEDGER_PUBLIC_URI` option.

Note that the previous code (uris starting with /) actually didn't make sense for red/blue.ilpdemo.org because the ledger there is running at a subdirectory: https://red.ilpdemo.org/ledger/

So `/health` would actually mean `https://red.ilpdemo.org/health` (wrong), not `https://red.ilpdemo.org/ledger/health` (correct). This code will always return the complete, correct URI.

Relative URIs are harder to handle for the client because they need to know how to resolve them, and support for relative URIs is only really mature in the HTML/browser context, not in a JSON/API context. (Unless you use JSON-LD, which may be a future improvement we want to look into.)

src/lib/notificationWorker.js
justmoon: Not really, will remove.

src/controllers/metadata.js
clark800: Makes sense, I didn't think about the issue with mounting at subdirectories.

PR #236, opened 2016-04-28
Provide notifications via websockets

justmoon:
Adds a new /accounts/:name/transfers endpoint which provides transfer notifications via the websocket protocol.

### DIFF
--- a/src/lib/notificationWorker.js
+++ b/src/lib/notificationWorker.js
@@ -2,6 +2,7 @@
 
 const _ = require('lodash')
 const co = require('co')
+const EventEmitter = require('events').EventEmitter
 const utils = require('./notificationUtils')
 const NotificationScheduler = require('five-bells-shared').NotificationScheduler
 const transferDictionary = require('five-bells-shared').TransferStateDictionary
@@ -27,8 +28,10 @@ function * findOrCreate (Notification, data) {
   return yield Notification.findWhere(data.where, options)
 }
 
-class NotificationWorker {
+class NotificationWorker extends EventEmitter {
   constructor (uri, log, Notification, Transfer, Subscription, Fulfillment, config) {
+    super()
+
     this.uri = uri
     this.log = log
     this.Notification = Notification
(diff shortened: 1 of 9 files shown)


justmoon:
Just looked at the test failures and after thinking about it some more, I noticed an error in thinking - I'm sending the notifications before the database transaction has returned, which is faster, but unsafe. The ledger may end up sending out a notification about a transfer state update that never ends up happening due to a database failure.

Please hold off on reviewing until I can fix this issue.

justmoon:
Ok, ready to review!

src/lib/notificationWorker.js
```diff
@@ -50,18 +53,38 @@ class NotificationWorker {
 
   * queueNotifications (transfer, transaction) {
     const affectedAccounts = _([transfer.debits, transfer.credits])
-      .flatten().pluck('account').map((account) => this.uri.make('account', account)).value()
+      .flatten().pluck('account').value()
     affectedAccounts.push('*')
+
+    const fulfillment = yield this.Fulfillment.findByTransfer(transfer.id, { transaction })
```
MatthewPhinney: We could save some time by only looking for the fulfillment if the transfer state is `executed` or `rejected`.

MatthewPhinney:
Out of curiosity, do we have any data regarding how this will affect performance?

src/lib/notificationWorker.js
justmoon: Good idea, will add.

justmoon:
@MatthewPhinney 
> Out of curiosity, do we have any data regarding how this will affect performance?

Not yet, but @nguyenripple is working on adding performance tests to CI.

Logically, a websocket frame is a single TCP packet as opposed to an HTTP request which opens up a whole new TCP connection. We are also avoiding at least two database writes (writing the notification intent and then deleting it) by dropping the guaranteed delivery requirement.

Anecdotally, I tried it in five-bells-demo with universal mode and it's visibly faster. Note that universal mode will see a much greater improvement than atomic, since this change speeds up both phases in universal and only the preparation phase in atomic.

Also note that the longer the payment path, the more of an improvement you'll see and in the demo we're dealing usually with two, three or even more hops, whereas our integration tests only go up to two hops currently.

clark800:
LGTM 👍

PR #238, opened 2016-05-02
[FIX] Allow scale=4

sentientwaffle:


emschwartz:
What happens if `LEDGER_AMOUNT_SCALE` is larger than the value hard-coded in the database creation?

sentientwaffle:
Then it will be broken, same as before. I needed scale=4 to work because the integration tests use it.

MatthewPhinney:
I know @lumberj put in #215 to allow the ledger to have configurable precision. Might be good for him to take a quick look. Also, I'm curious what broke the integration tests? Why do they need scale=4?

sentientwaffle:
The integration tests were written for scale=4. Fixing the ledger to (correctly) support it is simpler than rewriting the tests.

gip:
Why are we hardcoding the scale? Ideally scale should be coming from configuration when creating the tables too.

sentientwaffle:
I agree, but that's out of the scope of this PR.

clark800:
LGTM after tests pass

PR #242, opened 2016-05-10
[TASK] Add switch to disable notification signing (RILP-478)

MatthewPhinney:
cc @gip @nguyenripple


src/lib/notificationWorker.js
```diff
@@ -154,26 +154,31 @@ class NotificationWorker extends EventEmitter {
         }
       }
     }
-    // Sign notification
-    const algorithm = 'CC' // Crypto-condition signatures
-    let signedNotification
-    if (this.signatureCache[notification.id]) {
-      signedNotification = _.extend(notificationBody, { signature: this.signatureCache[notification.id] })
+
+    const signNotification = this.config.get('keys.notification_sign.must_sign')
```
clark800: Sounds like a method name, maybe `mustSignNotifcation` or just inline it?

src/lib/config.js
```diff
@@ -53,6 +53,9 @@ function parseFeaturesConfig () {
 }
 
 function parseKeysConfig () {
+  const isProduction = process.env.NODE_ENV === 'production'
+  const mustSignNotifications = Config.castBool(
+    Config.getEnv(envPrefix, 'NOTIFICATION_SIGN'), isProduction)
```
clark800: `SIGN_NOTIFICATIONS` sounds more natural

gip: We already have a `CONNECTOR_NOTIFICATION_VERIFY`config option though. We should be consistent.

gip:
LGTM

MatthewPhinney:
Discussed with @nguyenripple and we decided this is an unnecessary complication at this point. http notifications will always be signed, and ws notifications will never be signed. Notification signing no longer appears to be a performance bottleneck. If this changes, we can re-open this PR.

clark800:
> and ws notifications will never be signed

For non-`wss` websocket notifications, wouldn't they still have to be signed?

MatthewPhinney:
I just mean, in the current implementation, https notifications are signed and ws notifications are not. Adding signatures to non-wss notifications could be another task. But for this PR, I didn't see any benefit to introducing another flag to change http-notification signature behavior.

emschwartz:
+1 to adding ws notification signing

On Tue, May 10, 2016 at 2:05 PM, Matthew Phinney <notifications@github.com>
wrote:

> I just mean, in the current implementation, https notifications are signed
> and ws notifications are not. Adding signatures to non-wss notifications
> could be another task. But for this PR, I didn't see any benefit to
> introducing another flag to change http-notification signature behavior.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/interledger/five-bells-ledger/pull/242#issuecomment-218291157>
>



-- 
Evan Schwartz | Software Architect | Ripple
[image: ripple.com] <http://ripple.com>

PR #243, opened 2016-05-10
Switch to PBKDF2

justmoon:
Requires interledger/five-bells-shared#136

Removes password hashing from five-bells-ledger and uses the facility in five-bells-shared.

Makes password hashing asynchronous. Each password hashing iteration with 10 rounds of bcrypt takes around 60ms in my tests. This is for every request to the ledger and on the main thread. With this patch, the password hashing becomes asynchronous and is removed from the main thread. This reduces the total duration of the unit tests from 35s to about 8s on my machine. The number of "prepare transfer" requests per second increases from 60 req/s to about 110 req/s.

Note that the unit tests were using some rudimentary partial caching for password hashes, so the impact on a live system could be even larger, depending on the number of cores available.

This change does not modify the public API, however it does change the database schema and password hashing and therefore necessitates a database reset.

### DIFF
--- a/src/models/db/account.js
+++ b/src/models/db/account.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const bcrypt = require('bcrypt')
 const Model = require('five-bells-shared').Model
 const PersistentModelMixin = require('five-bells-shared').PersistentKnexModelMixin
 const uri = require('../../services/uriManager')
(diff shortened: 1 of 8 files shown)


justmoon:
Tests will pass after interledger/five-bells-shared#136 is merged and `package.json` updated.

emschwartz:
LGTM (once tests pass on CI)

lumberj:
> With this patch, the password hashing becomes asynchronous and is removed from the main thread.

How does this work? I think it still executes on the main thread, but on the next tick of the event loop

lumberj:
> How does this work? I think it still executes on the main thread, but on the next tick of the event loop

Oh I see, pbkdf2 is a native built-in

clark800:
LGTM after tests pass

src/models/db/account.js
```diff
@@ -44,10 +25,8 @@ class Account extends Model {
     if (data.balance) {
       data.balance = Number(data.balance)
     }
-    if (data.password) {
-      data.password_hash = hashPassword(data.password)
-      delete data.password
-    }
+
+    delete data.password_hash
```
MatthewPhinney: Why are we deleting the password hash?

lumberj: This is converting to external format. This is already deleted on line 67

MatthewPhinney: Oh, I see we delete the password in `setAccount` now. Might be a bit clearer to delete this in `convertToPersistent`.

justmoon: This is actually converting *from* external format. The reason I delete it is I want to prevent exploits where somebody can pass in a password hash. Passing in a password hash is a DoS vector because I can set the iterations to UINT32_MAX and basically make you do an infinite amount of work.

I'll add a comment to explain.

MatthewPhinney: That makes sense. The reason this initially jumped out at me is that we were deleting the hash, but not the password (I didn't see the code in `setAccount`), so I thought we would store the plaintext password in the db.

PR #245, opened 2016-05-11
fix: teardown database connection on exit

lumberj:


lumberj:
Confirmed that this works locally. I used an Oracle connection, which uses a connection pool

MatthewPhinney:
LGTM

src/app.js
```diff
@@ -1,7 +1,20 @@
...
+    console.error('Exiting: ', error)
+    process.exit(1)
+  }
+  process.exit(0)
+}
+
 module.exports = require('./lib/app')
 if (!module.parent) {
+  process.on('SIGINT', exitHandler)
+  process.on('uncaughtException', exitHandler)
```
clark800: Do we want this? I've always thought `process.on('uncaughtException'` was pretty risky.

lumberj: It's risky if you allow the process to live after this happens

lumberj: https://nodejs.org/api/process.html#process_warning_using_uncaughtexception_correctly

> Attempting to resume normally after an uncaught exception can be similar to pulling out of the power cord when upgrading a computer -- nine out of ten times nothing happens - but the 10th time, the system becomes corrupted.

clark800: Yeah, that's explaining why you don't want to use `process.on('uncaughtException'`. It's better to stick with the default behavior, which is just to exit, apparently without cleanup. The risk is that you end up suppressing the stacktrace and make it very hard to debug, which is what it looks like will happen with the current code.

lumberj: > The risk is that you end up suppressing the stacktrace and make it very hard to debug, which is what it looks like will happen with the current code.

I'll add a logger statement and log the stack trace

lumberj: Should be good now

src/app.js
```diff
@@ -1,7 +1,20 @@
 /* @flow */
 'use strict'
 
+// SIGINT and uncaughtException do not run cleanup
+// so we have an exitHandler to clean up the database connection
+function exitHandler (error) {
+  if (error) {
+    console.error(error.stacktrace)
```
clark800: You mean `error.stack`?

lumberj: My brain is 😴 

Let me fix that

justmoon:
Can you add a description? What is this for? Your exit handler just calls `process.exit()`, so I assume there is something else that is triggered by that?

src/app.js
justmoon: Definitely agree that this seems dangerous and hacky.

``` js
throw null
```

^ makes this PR exit with no error, no explanation and status code 0 (success).

Normal behavior is to print:

```
/opt/workspace/five-bells-ledger/test.js:1
(function (exports, require, module, __filename, __dirname) { throw null
                                                              ^
null
```

I'm not so much arguing that you need to handle `throw null` correctly, more illustrating why overriding the default exception handler is problematic. I'm trying to understand more about the problem that you're trying to solve, maybe there is a less hacky approach.

lumberj: >  I'm trying to understand more about the problem that you're trying to solve, maybe there is a less hacky approach.

This is the problem we're solving. SIGINT is sent by process manager to the node process when it wants to shut it down. Knex maintains a connection pool, which does not close automatically on a SIGINT. So we have to trap that signal, and exit the process, which forces the connection to close.
See:
https://github.com/tgriesser/bookshelf/issues/405

`throw null` is a bug in javascript it seems. Not sure how we'd get around that.

lumberj:
I'll add the description I gave in https://github.com/interledger/five-bells-ledger/pull/245#discussion_r62935916

lumberj:
I am going to remove the uncaught exception event handler

lumberj:
@justmoon how does the default exception handler handle uncaught exceptions?

src/app.js
justmoon: > Knex maintains a connection pool, which does not close automatically on a SIGINT.

Oh, so just call `knex.destroy()`.

If you don't call `process.exit()`, then you're not preventing the default handler you're just adding something to it. That should be pretty safe for both on-exit and on-uncaughtException. (They're [just standard EventEmitters](https://github.com/nodejs/node/blob/4d4cfb27ca7718c7df381ac3b257175927cd17d1/lib/internal/bootstrap_node.js#L268), so as long as your listener doesn't throw and doesn't `process.exit`, they'll do their thing after it returns.)

lumberj: > Oh, so just call knex.destroy().

I've tried this and it didn't work locally. I'll test again. 

> If you don't call process.exit(), then you're not preventing the default handler you're just adding something to it.

I don't think I understand what you mean by "default handler". SIGINT is sent to a process when a user presses ctrl+c. A process catches SIGINT to be able to gracefully shutdown. I don't think theres anything to do other than exiting after cleaning up in the event handler. There are no default handlers.

lumberj: I've just tested again locally.

1) Calling knex.destroy() in process.on('exit') doesn't do anything. The documentation states, "Therefore you must only perform synchronous operations in this handler." knex.destroy is async

2) If the process handles SIGINT and doesn't exit, it must be sent a SIGKILL (`kill -9`)

This must be sent a SIGKILL for it to exit, ctrl+c won't work.
```
process.on('SIGINT', () => {})
setTimeout(() => {}, 1000000)
```

lumberj: I think maybe the confusion lies in the interface Node provides for trapping signals. It would have made more sense for Node to call this `process.trap('SIGINT', callback)` --

http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_12_02.html
https://docs.python.org/3.5/library/signal.html#signal.signal

justmoon: Yeah I misremembered, I thought you had on-exit, not on-SIGINT. Long story short I think that exiting on SIGINT is totally fine - so if you take out the on-uncaughtException, then this LGTM.

PR #247, opened 2016-05-12
fix: db precision is too low if funding accounts > 1,000,000 balance

lumberj:
This change, 256758d, introduced a change in the scale (digits after
decimal place). To support amounts greater than 1,000,000 (6 digits) we
need more than 10 digits of precision (6 digits before + 4 digits after
decimal point).


clark800:
LGTM

PR #249, opened 2016-05-16
[TASK] Remove delete functionality (RILP-447)

MatthewPhinney:


emschwartz:
What happens if you post a new subscription with the same ID as one that was deleted before?

src/migrations/201602231400_initial.js
```diff
@@ -53,6 +53,7 @@ function createSubscriptionsTable (knex) {
     table.string('event')
     table.string('subject', 1024)
     table.string('target', 1024)
+    table.boolean('is_deleted').index().defaultTo(false)
```
clark800: I think you want a multi-column index with `id` and `is_deleted` right?

MatthewPhinney:
> What happens if you post a new subscription with the same ID as one that was deleted before?

The same thing that happens today if a user tries to create a subscription with a previously used UUID: the database throws a unique constraint violation. I can see how this could be confusing to users: they delete a subscription, but save the UUID and then try to create a new subscription with that UUID. I don't see a good reason for why they would do this, but I'm open to suggestion on how to handle this case.

src/migrations/201602231400_initial.js
MatthewPhinney: Yeah, if it's going to be indexed at all, it should probably be a multi-column index. An index on a boolean field doesn't really make sense, since so many values repeat. I believe Oracle offers bitmapped indices for such cases, but knex doesn't support adding index type for anything besides postgres. So, presumably, this defaults to some kind of b-tree in oracle and sql server. (I tried to capture some of this info in the JIRA ticket.)

emschwartz:
What's the point of not properly deleting subscriptions? I understand the general principle of the `is_deleted` flag rather than really deleting records, but subscriptions don't seem very important...

clark800:
> What's the point of not properly deleting subscriptions?

According to Abraham, enterprise DBAs will likely not be willing to give our software delete privileges.

emschwartz:
Gotcha, thanks for the explanation. That's a good reason

src/models/db/subscriptions.js
```diff
@@ -48,7 +49,7 @@ function * upsertSubscription (subscription, options) {
 }
 
 function * deleteSubscription (id, options) {
-  yield Subscription.destroy(_.assign({}, options, {where: {id: id}}))
+  return yield upsertSubscription({id, is_deleted: true}, options)
```
clark800: This will insert a row even if a subscription with this `id` never existed.

MatthewPhinney:
OK, there was a bit more subtlety here than I first realized, but I think this is now in a reviewable state. The behavior is as follows:

- subscriptions are created with `is_deleted=false`
- when a user deletes a subscription, we set `is_deleted=true` for this ID. If the ID doesn't exist, we [return a 404](https://github.com/interledger/five-bells-ledger/pull/249/files#diff-765463c6f194fb585a552d0ca4422dcaR230). This addresses @clark800's point about not creating new records when deleting subscriptions.
- we filter out subscriptions where `is_deleted=true`, so the user [gets a 404] (https://github.com/interledger/five-bells-ledger/pull/249/files#diff-765463c6f194fb585a552d0ca4422dcaR65) when requesting deleted subscriptions
- if the user tries to update or create a deleted subscription, we [return a 400](https://github.com/interledger/five-bells-ledger/pull/249/files#diff-765463c6f194fb585a552d0ca4422dcaR126). This addresses @emschwartz concern about db constraint violations.
- once a subscription is deleted, we [no longer send notifications](https://github.com/interledger/five-bells-ledger/pull/249/files#diff-6e15316ac94b638da1a3cf7c69ba2ee0R1065)

Additional feedback is welcome.

src/models/db/subscriptions.js
```diff
@@ -48,7 +55,11 @@ function * upsertSubscription (subscription, options) {
 }
 
 function * deleteSubscription (id, options) {
-  yield Subscription.destroy(_.assign({}, options, {where: {id: id}}))
+  const existingSubscription = yield getSubscription(id, options)
```
clark800: I think this should be in a transaction

MatthewPhinney: I think it is: `db.deleteSubscription(id, {transaction})`. As long as the transaction is passed in, it should be used throughout this method, right? I can add an `assert(options.transaction)` as in `_upsertSubscription`

MatthewPhinney:
@clark800 updated

clark800:
LGTM

PR #251, opened 2016-05-19
Use SQL constraint to check minimum balance limit

clark800:


src/migrations/201602231400_initial.js
```diff
@@ -1,42 +1,61 @@
...
+  return knex.schema.raw(['create table if not exists "accounts" (',
+    '"id" integer not null primary key autoincrement, ',
+    '"name" varchar(255), ',
+    '"balance" float check (balance >= minimum_allowed_balance), ',
+    '"connector" varchar(1024), ',
+    '"password_hash" varchar(1024), ',
+    '"public_key" text, ',
+    '"is_admin" boolean, ',
+    '"is_disabled" boolean, ',
+    '"fingerprint" varchar(255), ',
```
MatthewPhinney: Do you have to add an index for fingerprint?

src/migrations/201602231400_initial.js
```diff
@@ -1,42 +1,61 @@
 'use strict'
 
+function createTableIfNotExists (knex, tableName, callback) {
```
MatthewPhinney: Why is this needed?

src/models/transfers.js
```diff
@@ -256,17 +257,26 @@ function isAuthorized (transfer) {
...
 }
 
-function * processTransitionToPreparedState (transfer, accountBalances) {
+function * processTransitionToPreparedState (transfer, transaction) {
   if (transfer.state === transferStates.TRANSFER_STATE_PROPOSED && isAuthorized(transfer)) {
-    yield accountBalances.applyDebits()  // hold sender funds
+    try {
+      yield holds.holdFunds(transfer, transaction)  // hold sender funds
+    } catch (e) {
+      if (e.code === 'SQLITE_CONSTRAINT') {
```
MatthewPhinney: Is there any way we can get a more specific error? Presumably this won't work for oracle or sql server.

MatthewPhinney:
The logic is so much simpler now :smile:

src/migrations/201602231400_initial.js
clark800: This is to make it easier to dump the SQL schemas

src/migrations/201602231400_initial.js
clark800: Yeah, I'll fix that while writing the SQL schemas

src/models/transfers.js
clark800: I'm not sure, but yeah at the very least I need to make it not specific to SQLite

clark800:
Tests are passing! Ready for review

lumberj:
The commit title/message should be updated to reflect all the changes

test/helpers/db.js
```diff
@@ -27,12 +28,22 @@ const migrationTables = [
 // Only run migrations once during tests
 let init = false
 exports.init = function * () {
+  // TODO: Create dropTable fn and run before createTable
+  // To reset database
+  // Use LEDGER_DB_URI=oracle://ledger0_north:password@localhost/xe LEDGER_DB_SYNC=1 npm start
   if (init) return
+
+  if (knexConfig.client === 'strong-oracle') {
+    // oracle schemas must be loaded with "npm run setup-oracle"
```
lumberj: Make the logger log this message as a reminder?

test/helpers/db.js
```diff
@@ -27,12 +28,22 @@ const migrationTables = [
 // Only run migrations once during tests
 let init = false
 exports.init = function * () {
+  // TODO: Create dropTable fn and run before createTable
```
lumberj: I don't think this comment is necessary anymore (I had it in my branch as a reminder)

lumberj:
+1. Need to take a deeper look at the account balances changes

package.json
```diff
@@ -16,7 +16,9 @@
     "start-prof": "node --prof --logfile=${LEDGER_V8_LOGFILE:-v8.log} src/app.js",
     "lint": "eslint .",
     "test": "istanbul test -- _mocha",
-    "test-oracle": "LEDGER_UNIT_DB_URI='oracle://system:oracle@192.168.99.100:49161/' DYLD_LIBRARY_PATH=/opt/oracle/instantclient LD_LIBRARY_PATH=/opt/oracle/instantclient node node_modules/.bin/istanbul test -- _mocha",
+    "sqlplus": "LD_LIBRARY_PATH=/opt/oracle/instantclient DYLD_LIBRARY_PATH=/opt/oracle/instantclient /opt/oracle/instantclient/sqlplus system/oracle@localhost:49161/xe",
+    "setup-oracle": "echo exit | LD_LIBRARY_PATH=/opt/oracle/instantclient DYLD_LIBRARY_PATH=/opt/oracle/instantclient /opt/oracle/instantclient/sqlplus system/oracle@localhost:49161/xe @src/sql/strong-oracle/create.sql",
+    "test-oracle": "LEDGER_UNIT_DB_URI='oracle://system:oracle@localhost:49161/xe' DYLD_LIBRARY_PATH=/opt/oracle/instantclient LD_LIBRARY_PATH=/opt/oracle/instantclient node node_modules/.bin/istanbul test -- _mocha",
```
MatthewPhinney: Why did this change from `localhost`?

package.json
```diff
@@ -16,7 +16,9 @@
     "start-prof": "node --prof --logfile=${LEDGER_V8_LOGFILE:-v8.log} src/app.js",
     "lint": "eslint .",
     "test": "istanbul test -- _mocha",
-    "test-oracle": "LEDGER_UNIT_DB_URI='oracle://system:oracle@192.168.99.100:49161/' DYLD_LIBRARY_PATH=/opt/oracle/instantclient LD_LIBRARY_PATH=/opt/oracle/instantclient node node_modules/.bin/istanbul test -- _mocha",
+    "sqlplus": "LD_LIBRARY_PATH=/opt/oracle/instantclient DYLD_LIBRARY_PATH=/opt/oracle/instantclient /opt/oracle/instantclient/sqlplus system/oracle@localhost:49161/xe",
```
MatthewPhinney: So this starts a command line instance of sqlplus?

scripts/ci-test.sh
```diff
@@ -69,24 +72,23 @@ oracletest() {
...
+  npm run setup-oracle
+  npm run test-oracle
 }
 
 oneNode() {
   lint
   dockerBuild
   sqlitetestest
   integrationtest
-  mysqltest
```
MatthewPhinney: Was the decision to drop mysql support run past the five-bells team?

emschwartz: Yeah, wait, why is that being removed?

src/sql/pg/create.sql
```diff
@@ -0,0 +1,69 @@
...
+"transfer_id" char(36),
+"retry_count" integer,
+"retry_at" timestamp);
+
+create index notifications_retry_at_index on "notifications" ("retry_at");
+create index subscription_transfer
+  on "notifications" ("subscription_id", "transfer_id");
+
+
+create table if not exists "entries" (
```
MatthewPhinney: What happened to the `balance` field in this table?

src/sql/pg/create.sql
```diff
@@ -0,0 +1,69 @@
...
+"is_disabled" boolean,
+"fingerprint" varchar(255),
+"minimum_allowed_balance" float default 0);
+
+create unique index accounts_name_unique on "accounts" ("name");
+create index fingerprint on "accounts" ("fingerprint");
+
+
+create table if not exists "transfers" (
+"id" char(36) not null primary key,
```
MatthewPhinney: Are primary keys automatically indexed?

src/sql/strong-oracle/create.sql
```diff
@@ -0,0 +1,332 @@
...
+ALTER TABLE "transfers"
+  ADD CONSTRAINT  PK_TRANSFERS PRIMARY KEY ("id")
+/
+
+CREATE TABLE "entries"
+(
+  "id"                   INTEGER  NOT NULL ,
+  "transfer_id"          VARCHAR2(64) NULL ,
+  "account"              INTEGER NULL ,
+  "entry_balance"        NUMBER(32,16) DEFAULT  0  NOT NULL ,
```
MatthewPhinney: Here there is an `entry_balance`, but not in sqlite or postrgres.

src/sql/strong-oracle/create.sql
```diff
@@ -0,0 +1,332 @@
+BEGIN
```
MatthewPhinney: I'm not qualified to review this; I think Abraham and/or Nara should take a look. A quick overview of the `CREATE INDEX` and `ALTER TABLE` statements would be helpful though. For example, are `XPKL_SUBSCRIPTIONS` and `PK_ACCOUNTS` oracle naming conventions?

package.json
clark800: You mean _to_ `localhost`? It's because we are using new oracle setup instructions.

package.json
clark800: Yes

scripts/ci-test.sh
clark800: I don't think we have much choice since mysql does not enforce CHECK constraints, and the approach we used to fix this critical bug (money creation/destruction) depends on a CHECK constraint.

I'll post to the five-bells slack channel to see if anyone has any other ideas.

src/sql/pg/create.sql
scratch28: yes setting a column to PK will have a built in system generated index and unique constraint

package.json
scratch28: Ripple Connect uses strong oracle that leverage TNS, a better protocol for connecting to Oracle.   Something to emulate on next release

src/sql/pg/create.sql
clark800: Ah, I should add that back in now, thanks

src/sql/strong-oracle/create.sql
clark800: Should be `balance`, will change

src/sql/strong-oracle/create.sql
clark800: I believe these were Abraham's recommendations

clark800:
@lumberj @MatthewPhinney updated

src/lib/holds.js
```diff
@@ -0,0 +1,60 @@
...
+    throw new InsufficientFundsError('Sender has insufficient funds.', account)
+  })
+  /* eslint-enable */
+}
+
+function insertEntry (account, transferID, transaction) {
+  return transaction.select('id')
+    .from('accounts').where('name', account)
+    .then((rows) => {
+      const entry = {
```
MatthewPhinney: I think you want to add the balance in here too, right?

clark800: Ah, that must have gotten lost on another branch

clark800: Oh, I remember, it's because we haven't gotten it working for oracle!

src/sql/strong-oracle/create.sql
scratch28: Generally you want to test if the object exists first and then drop else move on 
not attempt to drop then raise error if it bombs.
```
declare
    table_count NUMBER :=0;  
BEGIN
    select count(*) into table_count from all_tables where table_name='YOUR_TABLE_NAME' ;
    if table_count >= 1 then
       EXECUTE IMMEDIATE 'DROP TABLE YOUR_TABLE_NAME      cascade constraints'; 
    end if;
end;
/

```

nararama: General standard is that all index script  should start with IDX_

lumberj: This is from the top-voted answer on stackoverflow: http://stackoverflow.com/questions/1799128/oracle-if-table-exists

scripts/ci-test.sh
scratch28: Check constraints in MySQL is handled as a trigger before insert and would have to rely on a fast query (usually in row math is best where you already have the base and delta values to evaluate).   That trigger can be ported to any db, but  since you plan to have multiple scripts for each supported flavor of db, it would just be a different methodology for the same function/logic.

src/sql/strong-oracle/create.sql
scratch28: since you are raising the error, do you wish for the script to stop mid flow?
if yes, then what you have is sufficient, if no then the lower ranked version (what I recommend to do) is better.

lumberj: I think that it only re-raises the error if it's not the error that's expected

lumberj: What we probably want is to move these to a `drop.sql` and instead use what you suggested here.

clark800: @scratch28 The stack overflow answer argues that one query is better than two, which is a small reason in this case, but so far I haven't heard any justifications for the other side, so I'll leave it as-is unless there is a justification to change.

clark800: > General standard is that all index script should start with IDX_

@nararama These names were @scratch28's suggestions, so please discuss with him. We have a separate task for renaming, so we can save those changes for a future pull request.

src/lib/holds.js
clark800: Let's leave that as a follow-up task because it's not easy to get it working in Oracle

lumberj:
LGTM

PR #252, opened 2016-05-24
[FIX] Transfer should be executed only once (RILP-558)

MatthewPhinney:


src/models/transfers.js
```diff
@@ -318,6 +318,13 @@ function * fulfillTransfer (transferId, fulfillmentUri) {
   const fulfillment = ConditionFulfillment.fromDataExternal(fulfillmentUri)
   fulfillment.setTransferId(transferId)
   const existingFulfillment = yield db.transaction(function * (transaction) {
+    // Set isolation level to avoid reading "prepared" transaction that is currently being
+    // executed by another request. This ensures the transfer can be fulfilled only once.
+    if (db.type === 'oracle' || db.type === 'postgres') {
```
clark800: Is this right? The client for postgres is "pg" and the client for oracle is "strong-oracle". Also, can you add the helper function as discussed earlier?

clark800: On second thought, a function isn't necessary, as long as we get the value from one place.

src/models/transfers.js
```diff
@@ -318,6 +318,12 @@ function * fulfillTransfer (transferId, fulfillmentUri) {
   const fulfillment = ConditionFulfillment.fromDataExternal(fulfillmentUri)
   fulfillment.setTransferId(transferId)
   const existingFulfillment = yield db.transaction(function * (transaction) {
+    // Set isolation level to avoid reading "prepared" transaction that is currently being
+    // executed by another request. This ensures the transfer can be fulfilled only once.
+    if (db.client === 'pg') {
```
MatthewPhinney: Only doing this in postgres for now, for reasons described in RILP-558.

clark800: I don't see the reason in that ticket, can you explain?

MatthewPhinney: I don't see it either ... I guess it wasn't saved properly. Long story short, I can't set the isolation level for mysql. I get `ERROR Error: ER_CANT_CHANGE_TX_CHARACTERISTICS: Transaction isolation level can't be changed while a transaction is in progress
`. It looks like it has to be set once on start up, rather than on a transaction by transaction basis.

I also can't set oracle, because we are using XE, which doesn't support this. Once we move to the standard edition, we should be able to set the isolation level on oracle.

I think this is OK for this ticket, since this is a blocker for RC, and has only been observed on postgres. I'm all for a most robust solution long term, but I don't want to delay this until later in the week, just to wait for Oracle standard edition.

MatthewPhinney: We may be able to use the `SESSION` keyword for mysql.

MatthewPhinney: @clark800 I updated to include mysql. Oracle still won't work until we move away from XE.

src/models/transfers.js
```diff
@@ -318,6 +318,11 @@ function * fulfillTransfer (transferId, fulfillmentUri) {
   const fulfillment = ConditionFulfillment.fromDataExternal(fulfillmentUri)
   fulfillment.setTransferId(transferId)
   const existingFulfillment = yield db.transaction(function * (transaction) {
+    // Set isolation level to avoid reading "prepared" transaction that is currently being
+    // executed by another request. This ensures the transfer can be fulfilled only once.
+    if (db.client === 'pg' || db.client === 'strong-oracle') {
+      yield transaction.raw('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE')
+    }
```
MatthewPhinney: @clark800 I updated this to include oracle. The issue was due to a semi-colon on the end of the SQL string. I removed mysql, since you said your script doesn't support it. I changed the isolation level to `serializable` for postgres because oracle doesn't support `repeatable read`. It only support `serializable` and `read committed`. See the [docs](https://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_10005.htm). I also changed postgres to `serializable` because I think it's confusing to have different isolation levels depending on the database.

clark800:
LGTM

PR #253, opened 2016-05-24
Add instructions for testing to README

MatthewPhinney:


README.md
```diff
@@ -53,7 +53,6 @@ With the following configuration options set as environment variables:
 * `LEDGER_ADMIN_USER` (default: `'admin'`) The admin account's username (an admin user can create/modify accounts).
 * `LEDGER_ADMIN_PASS` (default: none) The admin account's password.
 * `LEDGER_ADMIN_FINGERPRINT` (default: none) The admin account's TLS certificate fingerprint if using TLS Client Certificate Auth.
-* `LEDGER_UNIT_DB_URI` (default `'sqlite://'`) Database to use for running unit tests. Configured separately for safety.
```
clark800: Why is this removed?

MatthewPhinney: This is in the section on running the server, so setting this environment variable here won't do anything. For clarity, I wanted to put it in its own section on testing, since I've often overlooked it in the readme, and found myself looking through the code to figure out how to set the test database. It seems clearer to have a separate place for this, rather than having it sandwiched between `LEDGER_ADMIN_FINGERPRINT` and `LEDGER_AUTH_BASIC_ENABLED`, neither of which have anything to do with the database used.

clark800: Maybe we could add a note to the LEDGER_DB_URI line about the LEDGER_UNIT_DB_URI env var

clark800:
LGTM

PR #254, opened 2016-05-31
Revert ws dependency version due to conflict with koa-websocket depen…

clark800:
…dency

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,6 @@
     "sinon-chai": "^2.7.0",
     "spec-xunit-file": "0.0.1-3",
     "supertest": "^1.1.0",
-    "ws": "^1.1.0"
+    "ws": "^0.8.0"
   }
 }


emschwartz:
Is there a specific bug the version conflict causes? Since the ws update to [`v1.0.1` fixes a DoS vulnerability](https://github.com/websockets/ws/releases/tag/1.0.1) it would be good to update to this ASAP

clark800:
Yeah, the publish script fails. I don't think DoS is a concern because this is a devDependency.

emschwartz:
You're right that ws is only a devDependency but there is a DoS concern because the ledger uses `koa-websocket`, which depends on a DoS-able version of `ws`. I just opened an [issue](https://github.com/kudos/koa-websocket/issues/11) there but until that's updated I believe this vulnerability will affect the ledger.

gip:
LGTM

clark800:
@emschwartz That's true, thanks for opening the issue

PR #257, opened 2016-06-03
Remove Account model and rename database table and columns for accounts

clark800:


src/controllers/accounts.js
```diff
@@ -112,8 +112,10 @@ function * putResource () {
   const name = this.params.name
   request.validateUriParameter('name', name, 'Identifier')
   const account = this.body
-  request.assert.strictEqual(account.name, name.toLowerCase(),
-    'Account name must match the one in the URL')
+  if (account.name) {
+    request.assert.strictEqual(account.name.toLowerCase(), name.toLowerCase(),
```
MatthewPhinney: Are you lower casing here to allow case insensitive account naming?

src/lib/holds.js
```diff
@@ -3,7 +3,7 @@
 const InsufficientFundsError = require('../errors/insufficient-funds-error')
 
 function adjustBalance (account, amount, transaction) {
-  const updateSQL = 'UPDATE "accounts" SET "balance" = "balance" + ? WHERE "name" = ?'
+  const updateSQL = 'UPDATE "L_ACCOUNTS" SET "BALANCE" = "BALANCE" + ? WHERE "NAME" = ?'
```
MatthewPhinney: Can you remind me why this was raw SQL? Now that we have the persistence layer in db/accounts, it seems like we could add a method to that file that uses knex.

src/models/accounts.js
```diff
@@ -8,15 +8,93 @@ const db = require('./db/accounts')
...
+      data.minimum_allowed_balance = Number(data.minimum_allowed_balance)
+    }
+  }
+
+  return data
+}
+
+function convertToExternal (data) {
+  data = _.cloneDeep(data)
+  data.id = uri.make('account', data.name.toLowerCase())
```
MatthewPhinney: I see that `name` is lowercased here, so I think you can disregard my comment in `src/controllers/accounts.js`.

src/models/accounts.js
```diff
@@ -8,15 +8,93 @@ const db = require('./db/accounts')
 const hashPassword = require('five-bells-shared/utils/hashPassword')
 const NotFoundError = require('five-bells-shared/errors/not-found-error')
 const UnauthorizedError = require('five-bells-shared/errors/unauthorized-error')
+const InvalidBodyError = require('five-bells-shared/errors/invalid-body-error')
+const uri = require('../services/uriManager')
+const validator = require('../services/validator')
+
+function getPublicData (data) {
```
MatthewPhinney: What do you think about using the convention `_getPublicData` to mark this as private?

src/models/db/accounts.js
```diff
@@ -1,36 +1,121 @@
...
+  return getTransaction(options)
+    .from(TABLE_NAME).select()
+    .then((results) => results.map(convertFromPersistent))
 }
 
 function * getConnectorAccounts (options) {
-  const jsonAccounts = yield Account.findAll(options)
-  return jsonAccounts.filter((account) => account.connector)
-    .map(Account.fromDatabaseModel.bind(Account))
+  return (yield getAccounts(options))
```
MatthewPhinney: Not an issue for this this PR (since this is a refactor), but it might be more efficient if we query the db to return only connector accounts, rather than filtering after we get the response from the db.

src/models/db/accounts.js
```diff
@@ -1,36 +1,121 @@
...
+  return getTransaction(options)
+    .from(TABLE_NAME).select().where(key, value)
+    .then((results) => results.map(convertFromPersistent))
+    .then((results) => {
+      if (results.length === 1) {
+        return results[0]
+      } else if (results.length === 0) {
+        return null
+      } else {
+        assert(false)
```
MatthewPhinney: This is just a check for a situation that should never happen? Under what circumstances would this happen?

src/models/db/accounts.js
```diff
@@ -1,36 +1,121 @@
...
+  return getAccountByKey('FINGERPRINT', fingerprint, options)
+}
+
+function * updateAccount (account, options) {
+  return getTransaction(options)(TABLE_NAME)
+    .update(convertToPersistent(account))
+    .where('NAME', account.name)
+}
+
+function * insertAccount (account, options) {
```
MatthewPhinney: I had been following the previous convention of calling this method `bulkInsert`. I'm fine with this name, but we should be consistent in both places. Incidentally, do you think we need "Account" in the method names, since these are all in the `db/accounts` file? Could do `get`, `getAll`, `getByFingerprint`, etc. I don't have a strong preference, but something to think about.

src/models/db/accounts.js
```diff
@@ -1,36 +1,121 @@
...
+  return getTransaction(options)
+    .insert(convertToPersistent(account))
+    .into(TABLE_NAME)
 }
 
-function * getAccount (name, options) {
-  return yield Account.findByName(name, options)
+function * insertAccounts (accounts, options) {
```
MatthewPhinney: I had been following the previous convention of calling this method `bulkInsert`. I'm fine with this name, but we should be consistent in both places. Incidentally, do you think we need "Account" in the method names, since these are all in the `db/accounts` file? Could do `get`, `getAll`, `getByFingerprint`, etc. I don't have a strong preference, but something to think about.

src/controllers/accounts.js
clark800: This was to make the tests pass, I'm not sure exactly what we want here, so I went with what the tests were expecting.

src/lib/holds.js
clark800: knex doesn't support the `balance = balance + ?` part as far as I can tell

src/models/accounts.js
clark800: There's no need for that because the exports object enforces that it is private to this file

src/models/db/accounts.js
clark800: I noticed that too, but wanted to keep the changes minimal

src/models/db/accounts.js
clark800: My style has usually been to make function names understandable without the context of the filename because when you require it somewhere else you would have to change the name if it was just `get`, which means you have some extra complexity and can't use the ES6 curly bracket notation etc. `bulkInsert` would be inconsistent with this.

clark800:
@MatthewPhinney updated

src/models/db/accounts.js
MatthewPhinney: That's a good point, but bear in mind that the `convertToExternal`, `convertFromPersistent`, etc. methods have the same name across multiple tables/models. If you want to be totally consistent, those should probably be renamed to `convertToAccountPersistent`, etc.

MatthewPhinney:
LGTM

src/models/db/accounts.js
clark800: Those are module-private functions so the naming doesn't matter as much

PR #258, opened 2016-06-06
fix: use numeric not float type for postgres balances

lumberj:


lumberj:
https://www.postgresql.org/docs/9.5/static/datatype-numeric.html

clark800:
LGTM

PR #259, opened 2016-06-06
More logging with sqlplus

gip:


src/lib/db.js
```diff
@@ -26,7 +26,8 @@ function executeSQLPlus (sqlFilepath) {
     const host = connection.host
     const port = connection.port
     const database = connection.database
-    const login = user + '/' + password + '@' + host + ':' + port
+    // TODO: Oracle connection string passed as configuration to ledger
+    const login = user + '/' + password + '@' + host + '/xe:' + port
```
clark800: This conflicts with the documentation: https://docs.oracle.com/cd/B19306_01/server.102/b14357/qstart.htm#i1055567

gip: Yeah, but the amazing thing is that actually solved the connection issue. I found another way to do that so this change will be undone in next commit.

src/lib/db.js
```diff
@@ -26,6 +26,7 @@ function executeSQLPlus (sqlFilepath) {
     const host = connection.host
     const port = connection.port
     const database = connection.database
+    // TODO? Oracle connection string passed as configuration to ledger
```
clark800: This is already the way it is, isn't it?

src/lib/db.js
```diff
@@ -35,6 +36,8 @@ function executeSQLPlus (sqlFilepath) {
     const command = '/opt/oracle/instantclient/sqlplus'
     const args = [url, '@' + sqlFilepath]
     const process = spawn(command, args, {env})
+    process.stdout.on('data', (data) => console.log(data.toString()))
```
clark800: I don't know if we always want to show all this; could we send it to the debug logger so it is only shown in debug mode?

MatthewPhinney:
What's the status of this PR?

PR #263, opened 2016-06-09
Remove Notification model and rename database table and columns for n…

clark800:
…otifications


src/lib/notificationWorker.js
```diff
@@ -18,35 +17,40 @@ const convertToExternalTransfer = require('../models/converters/transfers')
   .convertToExternalTransfer
 const getAffectedSubscriptions = require('../models/db/subscriptions')
   .getAffectedSubscriptions
+const getMatchingNotification = require('../models/db/notifications')
+  .getMatchingNotification
+const notificationDAO = require('../models/db/notifications')
```
lumberj: What is DAO?

clark800: https://en.wikipedia.org/wiki/Data_access_object

clark800: Not to be confused with "The DAO" https://daohub.org/ or "Dao" https://en.wikipedia.org/wiki/Tao or ... :)

lumberj:
LGTM

PR #264, opened 2016-06-13
Remove Fulfillment model and rename database table and columns for fu…

clark800:
…lfillments and reduce duplication in database layer


src/lib/holds.js
```diff
@@ -1,34 +1,35 @@
...
+const entries = require('../models/db/entries')
 
-function adjustBalance (account, amount, transaction) {
-  const updateSQL = 'UPDATE "L_ACCOUNTS" SET "BALANCE" = "BALANCE" + ? WHERE "NAME" = ?'
+function adjustBalance (accountName, amount, transaction) {
   /* eslint-disable handle-callback-err */
-  return transaction.raw(updateSQL, [amount, account]).catch((error) => {
-    throw new InsufficientFundsError('Sender has insufficient funds.', account)
+  return accounts.adjustBalance(accountName, amount, {transaction})
+  .catch((error) => {
```
MatthewPhinney: What do you think about logging this error? It's possible that there could be some database integrity issue (account doesn't exist, etc). I think that should also allow you to remove the `eslint-disable`.

src/models/db/utils.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  function insertAll (data, transaction) {
+    return (transaction || knex).insert(data.map(convertToPersistent))
+      .into(tableName)
+  }
+
+  function _upsert (data, where, transaction) {
+    assert(transaction, 'transaction is required for upsert')
+    return selectOne(where, transaction).then((existing) => {
```
MatthewPhinney: The `upsert` method uses `existed`. Might make sense to also name this `existed` for consistency.

src/controllers/accounts.js
```diff
@@ -115,6 +115,7 @@ function * putResource () {
   if (account.name) {
     request.assert.strictEqual(account.name.toLowerCase(), name.toLowerCase(),
       'Account name must match the one in the URL')
+    account.name = account.name.toLowerCase()
```
MatthewPhinney: So account names are now case insensitive?

src/models/db/utils.js
clark800: `existed` is a boolean, `existing` is an object

src/controllers/accounts.js
clark800: I don't know, but tests pass so if they aren't supposed to be we should add a test case for that

src/models/converters/entries.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const _ = require('lodash')
+
+function convertToInternalEntry (data) {
```
MatthewPhinney: I don't see these conversion methods used. I guess they're just included for completeness?

clark800: Yeah, they were in the model before, so I kept them around. We could delete them though.

src/models/db/utils.js
```diff
@@ -0,0 +1,85 @@
...
+        throw new InvalidModificationError(
+          'Already deleted, please use a new ID: ' + JSON.stringify(where))
+      }
+      const execute = existing
+        ? update(data, where, transaction) : insert(data, transaction)
+      return execute.then(() => Boolean(existing))
+    })
+  }
+
+  function withTransaction (callback) {
```
MatthewPhinney: Does this method need to be returned (at the end of the function)? I don't see it used externally.

clark800: Not currently, but it could be used in a future refactoring instead of the `co`-based version currently in use.

MatthewPhinney:
LGTM

src/lib/holds.js
clark800: I think it would be confusing to see an error in the logs when it's actually correct behavior. If you can find a way to detect just the expected error, that would be good.

clark800:
CirlceCI was failing yesterday, but I just ran three times in CircleCI and it passed every time. Abraham says we can merge because "the error you ran into can be ignored and all you have to do (if you ever run into again) is close off all existing connections and rerun your script".

PR #265, opened 2016-06-13
Add mysql create.sql

vhpoet:


src/sql/mysql/create.sql
```diff
@@ -0,0 +1,68 @@
+create table if not exists L_ACCOUNTS (
+ACCOUNT_ID int not null auto_increment primary key,
+NAME varchar(255),
+BALANCE numeric(32,16) check (BALANCE >= MINIMUM_ALLOWED_BALANCE),
```
clark800: MySQL does not enforce check constraints, which is why we removed MySQL support

justmoon:
Would `UNSIGNED` work instead? http://dev.mysql.com/doc/refman/5.7/en/numeric-type-attributes.html

clark800:
> Would UNSIGNED work instead?

Probably not, it looks like that just clamps the value to zero rather than generating an error: http://stackoverflow.com/a/8794634

Also, currently we are not shifting the balance so the balance value in the database can still go negative.

vhpoet:
Closing for now, until we get a better idea on how to proceed with this.

PR #267, opened 2016-06-15
[FIX] Default account balance to zero (RILP-595)

MatthewPhinney:


MatthewPhinney:
Actually, we may want this to go into `ilp-server-alpha`. What do you think @clark800 @lumberj @gip?

clark800:
I think we want all bugfixes in alpha

MatthewPhinney:
I will make this PR to the alpha branch.

PR #268, opened 2016-06-15
[FIX] Default balance to zero (RILP-595)

MatthewPhinney:


MatthewPhinney:
I haven't yet reproduced RILP-595 on `ilp_server_alpha`. I'm waiting to hear which version of the ledger was being used when this bug was discovered.

MatthewPhinney:
I added a couple test cases to catch this, but the only code change is to set the balance to 0 on account creation, if balance is unspecified. As far as I can tell, this bug was introduced [here](https://github.com/interledger/five-bells-ledger/commit/70361d0b93eabd631d02d887cdec465bf257b761). @clark800 let me know if you have any ideas on a better way to fix it.

src/models/db/accounts.js
```diff
@@ -30,6 +30,9 @@ function * _upsertAccount (account, options) {
     existingAccount.setData(account)
     yield existingAccount.save(options)
   } else {
+    if (!account.balance) {
+      account.balance = 0
```
clark800: We should avoid mutating function parameters, it leads to very subtle bugs

src/sql/pg/create.sql
```diff
@@ -1,7 +1,7 @@
 create table if not exists "accounts" (
 "id" serial not null primary key,
 "name" varchar(255),
-"balance" float check (balance >= minimum_allowed_balance),
+"balance" float default 0 check (balance >= minimum_allowed_balance),
```
clark800: Can you add `not null` constraints too?

MatthewPhinney: Yeah, I'd like to, but I'm getting some weird behavior when inserting multiple rows without balances. For some reason knex fills in the missing balance with `undefined` and the db throws a not null constraint, rather than filling in with the default value 0. This doesn't seem to be a problem when inserting one record at a time. Still looking into the best way to handle this.

clark800: To use the default you have to make sure that the column is not specified in the insert query at all

MatthewPhinney: Looks like a knex issue: https://github.com/tgriesser/knex/issues/662. Still looking into how to make it work.

src/models/db/accounts.js
```diff
@@ -30,7 +30,11 @@ function * _upsertAccount (account, options) {
     existingAccount.setData(account)
     yield existingAccount.save(options)
   } else {
-    yield Account.create(account, options)
+    if (!account.balance) {
+      yield Account.create(_.merge({}, account, {balance: 0}), options)
+    } else {
+      yield Account.create(account, options)
+    }
```
MatthewPhinney: I'm having trouble getting the modified db script to work with knex for several reasons.

1. knex v. 0.10 won't properly insert default values, so I can't set `default 0` for the `balance` column in the `accounts` table in the db script
2. knex v. 0.11 fixes the above issue, but breaks oracle: `Knex:warning - Pool2 - Error: ORA-00923: FROM keyword not found where expected`. The root cause is unknown, and could take some time to track down.
3. Even if the above two issues were fixed, we would still need this code for sqlite, since sqlite doesn't support inserting default values.

Since this is a hotfix for the release branch, I'm inclined to take the easy route and go with the above code, so that I have time to look into the oracle failures on knex v. 11 in more detail. What do you think @clark800?

clark800: This looks fine. So we aren't able to add the "not null" constraint?

MatthewPhinney: The problem with trying to enforce a not null constraint is that knex tries to insert `null` as the `default` on bulk inserts, so the db throw an error. We could work around this, since the only place we do bulk inserts is tests, but I wanted to keep the code changes as minimal as possible. I do think that setting a `not null` constraint is definitely a good approach long term, but that change could go into `master`.

clark800: It should be pretty easy to convert the bulk insert into a loop of single inserts right? I think one of the tables already works that way.

MatthewPhinney: It is not difficult, but that logic is in five-bells-shared, so it would require releasing a new version of that component, then including it here. I guess I can do that for now, since that would allow us to put in the not null constraint. When we merge this to master, the db code will be in this repo, so it will be more straightforward.

clark800: What code in five-bells-shared are you referring to? I thought this code in five-bells-ledger was what we were talking about: https://github.com/interledger/five-bells-ledger/blob/master/test/helpers/db.js#L39

MatthewPhinney: This branch doesn't yet have that code. It's still using the old models: https://github.com/interledger/five-bells-ledger/blob/bb3c60570f5d3030fa0df3ecd6e5971a6c6e33e9/test/helpers/db.js#L66. So it's here: https://github.com/interledger/five-bells-shared/blob/511482962dfccd0c5c9b57a57734edc2faac4f39/lib/persistent-knex-model-mixin.js#L117.

clark800: Oh, I see

MatthewPhinney:
requires https://github.com/interledger/five-bells-shared/pull/147

MatthewPhinney:
Closing as this is not required for alpha. I will make this PR to the master branch.

PR #270, opened 2016-06-16
Add L_LU_REJECTION_REASON lookup table

clark800:


src/models/db/transfers.js
```diff
@@ -31,6 +36,11 @@ function convertFromPersistent (data) {
   if (data.rejected_at) {
     data.rejected_at = new Date(data.rejected_at)
   }
+  if (!isNil(data.rejection_reason_id)) {
```
MatthewPhinney: Is this just a safety check? Or do some dbs return `undefined` and others return `null`?

src/sql/pg/create.sql
```diff
@@ -24,14 +24,33 @@ CREATE INDEX "L_XIE_FINGERPRINT" ON "L_ACCOUNTS"
...
+CREATE TABLE IF NOT EXISTS "L_LU_REJECTION_REASON" (
+  "REJECTION_REASON_ID"   INTEGER NOT NULL,
+  "NAME"                  CHARACTER VARYING(10) NOT NULL,
+  "DESCRIPTION"           CHARACTER VARYING(255) NULL
+);
+
+CREATE INDEX "L_XPK_LU_TRANSFERS_REJECTION_R" ON "L_LU_REJECTION_REASON"
+  ("REJECTION_REASON_ID" ASC);
+ALTER TABLE "L_LU_REJECTION_REASON" ADD CONSTRAINT
+  "L_PK_LU_TRANSFERS_REJECTION_RE"
```
MatthewPhinney: Should the suffix be `_R` for consistency?

src/models/db/transfers.js
clark800: Checking for `null` should suffice, I'll make that change

src/sql/pg/create.sql
clark800: I don't know, it's just getting truncated to 30 characters

MatthewPhinney: OK, I guess that makes sense. So the convention is to use the table name after the index/constraint name and just truncate at 30 chars.

MatthewPhinney:
LGTM

PR #272, opened 2016-06-20
Default balance to 0 in db (RILP-595)

MatthewPhinney:


test/accountSpec.js
```diff
@@ -81,7 +82,15 @@ describe('Accounts', function () {
       yield this.request()
         .get('/accounts')
         .auth('admin', 'admin')
-        .expect([account1, account2, account3, account4, account5, account6])
+        .expect((res) => {
+          const sortedResponse = _.sortBy(res.body, (account) => {
```
MatthewPhinney: Insertion order is no longer guaranteed, so we need to sort before we compare the arrays.

src/sql/pg/create.sql
```diff
@@ -1,7 +1,7 @@
 CREATE TABLE IF NOT EXISTS "L_ACCOUNTS" (
   "ACCOUNT_ID" SERIAL NOT NULL,
   "NAME" CHARACTER VARYING(255) NOT NULL,
-  "BALANCE" DECIMAL(32,16) NOT NULL CHECK
+  "BALANCE" DECIMAL(32,16) DEFAULT 0 NOT NULL CHECK
     ("BALANCE" >= "MINIMUM_ALLOWED_BALANCE"),
```
gip: Shouldn't `MINIMUM_ALLOWED_BALANCE` column be `NOT NULL` too?

MatthewPhinney: Probably. I was trying to figure out if this was captured in the db schema that Chris has been implementing. I can't find any info on it, so I'll add that constraint now.

test/accountSpec.js
gip: Couldn't the `noBalance` account be included in this test too?

src/sql/pg/create.sql
MatthewPhinney: Actually, I remembered that `MINIMUM_ALLOWED_BALANCE` must be nullable. The reason is, we use `NULL` in the db to correspond to `-infinity` (this is how we allow the balance to go negative without limit). See [here](https://github.com/interledger/five-bells-ledger/blob/77bdc25bc98f11a0bde6a7c472e5244a942b3fdc/src/models/db/accounts.js#L40).

gip: Cool. Having a comment somewhere in the code about that would help :)

gip:
LGTM

test/accountSpec.js
MatthewPhinney: It could, but then we wouldn't be able to create that account using the RESt api. This test checks direct insertion into the db, using `dbHelper.addAccounts`, while it `should default the balance to 0` tests using the RESt api. I think it's safest to test with the RESt api, since that's the endpoint we are exposing.

src/sql/pg/create.sql
MatthewPhinney: I added a comment.

PR #273, opened 2016-06-21
[FIX] Don't reset is_admin flag when account is updated

MatthewPhinney:


MatthewPhinney:
We were defaulting the `is_disabled` and `is_admin` flags to `false` in the code. This means that if a user updated an account with one of those flags already set to `true`, but did not explicitly specify the flag in their JSON request, it would be reset to `false`.

src/models/db/accounts.js
```diff
@@ -43,8 +41,12 @@ function convertToPersistent (data) {
       data.minimum_allowed_balance = Number(data.minimum_allowed_balance)
     }
   } // Otherwise the db defaults minimum_allowed_balance to 0
```
lumberj: Is this comment relevant to the code below it?

MatthewPhinney: No. I added that because @gip wanted a comment for #272 on how the minimum balance is handled in the db.

lumberj:
LGTM

PR #277, opened 2016-06-29
Breaking: Refactor notifications

gip:
* Removed http notifications and associated db tables / logic
* Client is able to ask for notification being resend
* Associated PR: https://github.com/interledger/five-bells-receiver/pull/2


mwshortt:
It seems like this is transferring robust notification responsibility from the ledger to the client, i.e. if RC is down and misses a ledger notification the ledger will not retry and instead it is up to RC to request that the notification be resent using the transfer id. Is that correct?

gip:
@mwshortt Absolutely. The ledger will not support robust notifications through http anymore. It's up to the client to implement a retry logic in case notification were missed. The ledger offers an easy way to ask for a notification to be resent.

In the future, we may reimplement robust notification through AMQP or a similar protocol.

test/accountSpec.js
```diff
@@ -632,6 +630,86 @@ describe('Accounts', function () {
...
+        .send(transfer)
+        .expect(201)
+        .expect(validator.validateTransfer)
+        .end()
+
+      // TODO: Is there a more elegant way?
+      yield timingHelper.sleep(50)
+
+      const transferId = transfer.id.match('http://localhost/transfers/(.+)')[1]
+      this.socket.send(transferId)
```
emschwartz: I think you should send some kind of request type, not just the transferId (in case you want to add other types of messages in the future):

```js
{
  type: 'request_notification',
  id: '...'
}
```

or something like that

test/putTransferSpec.js
```diff
@@ -1018,103 +1016,7 @@ describe('PUT /transfers/:id', function () {
         .end()
     })
 
-  /* Subscriptions */
-  it('should trigger subscriptions', function * () {
```
emschwartz: There should be a test to make sure notifications are sent out via websockets

test/fulfillmentSpec.js
```diff
@@ -331,162 +326,4 @@ describe('PUT /fulfillment', function () {
...
-      .expect(201)
-      .expect(this.executionConditionFulfillment)
-      .expect(validator.validateFulfillment)
-      .end()
-    yield notificationWorker.processNotificationQueue()
-
-    notificationExecuted.done()
-  })
-
-  it('should trigger subscriptions when notification is cancelled', function * () {
```
emschwartz: This test should be recreated with websockets

test/fulfillmentSpec.js
```diff
@@ -331,162 +326,4 @@ describe('PUT /fulfillment', function () {
     expect((yield Account.findByName('alice')).balance).to.equal(100)
     expect((yield Account.findByName('bob')).balance).to.equal(0)
   })
-
-  it('should trigger subscriptions when notification is executed', function * () {
```
emschwartz: This test should be recreated with websockets

src/services/notificationHandler.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
+
+const NotificationHandler = require('../lib/notificationHandlerWebsocket')
```
emschwartz: Out of curiosity, what's the point of changing `notificationWorker` to `notificationHandler`? (Handler kind of sounds to me like something that receives notifications, not something that sends them out)

src/models/notifications.js
```diff
@@ -1,59 +0,0 @@
...
-const Notification = require('./db/notification').Notification
-const Transfer = require('./db/transfer').Transfer
-const Fulfillment = require('./db/conditionFulfillment').ConditionFulfillment
-const subscriptions = require('./db/subscriptions')
-const subscriptionUtils = require('../lib/subscriptionUtils')
-const transferDictionary = require('five-bells-shared').TransferStateDictionary
-
-const transferStates = transferDictionary.transferStates
-
-function * getNotification (subscriptionId, notificationId, requestingUser) {
```
emschwartz: Might it still be useful to be able to `GET` notifications (as an alternative to requesting it via the websocket channel)?

src/models/db/notification.js
```diff
@@ -1,18 +0,0 @@
-'use strict'
-
-const Model = require('five-bells-shared').Model
-const PersistentModelMixin = require('five-bells-shared').PersistentKnexModelMixin
-const knex = require('../../lib/knex').knex
-
-class Notification extends Model {
```
emschwartz: Why isn't this needed anymore?

src/services/notificationHandler.js
gip: Everything is event driven so there is no worker anymore. Worker gives the notion that a background process is doing something, which is not the case anymore. A handler just does the right thing with your notifications.

src/lib/notificationWorker.js
```diff
@@ -1,192 +0,0 @@
...
-          execution_condition_fulfillment: fulfillment.getDataExternal()
-        }
-      } else if (transfer.state === transferStates.TRANSFER_STATE_REJECTED) {
-        notificationBody.related_resources = {
-          cancellation_condition_fulfillment: fulfillment.getDataExternal()
-        }
-      }
-    }
-    // Sign notification
-    const algorithm = 'CC' // Crypto-condition signatures
```
emschwartz: Are you sure we want to get rid of notification signing? Is that deemed unnecessary because we're trusting the `wss` connection?

src/lib/notificationHandlerWebsocket.js
```diff
@@ -0,0 +1,52 @@
+'use strict'
```
MatthewPhinney: Including `Websocket` in the the filename seems unnecessary. We don't include other types of notifications anymore, right?

src/controllers/notifications.js
```diff
@@ -1,67 +0,0 @@
...
- *        "state": "executed"
- *      },
- *      "related_resources": {
- *        "execution_condition_fulfillment": "cf:0:_v8"
- *      }
- *    }
- *
- * @apiUse NotFoundError
- * @apiUse InvalidUriParameterError
- * @apiUse UnauthorizedError
```
emschwartz: Is there a way to add the apidoc comments back but for the websocket connection?

test/putTransferSpec.js
gip: Is this not sufficient: https://github.com/interledger/five-bells-ledger/blob/refactor_notifications/test/accountSpec.js#L603 ?

src/models/transfers.js
```diff
@@ -461,10 +453,26 @@ function * getFulfillment (transferId) {
   return fulfillment.getDataExternal()
 }
 
+function * resendNotification (transferId) {
```
MatthewPhinney: Do you see any benefit to emitting an event to indicate there was an error (in the case that the transfer is not found, for example)? My understanding is that this method will be used when a client (such as RC) requests the state of a particular transfer. If the client has a transfer that the ledger does not, that indicates a pretty serious data integrity issue.

src/models/db/notification.js
gip: The `notifications` and `subscriptions` tables are basically gone. Once we update a transfer, the ledger will send a notification through websocket and then forget about it. By design, these notifications aren't robust. It's up to the client to send a request for a notification to be resent if it didn't get it.

Robust notifications in the ledger are very expensive in terms of transactions (over more than one table). We may want to reimplement a robust notifications using AMQP or something like that. That would move some complexity to another specialized subsystem (a queue).

src/lib/notificationWorker.js
gip: Yes. I thought that since `wss` connection are authenticated so that shouldn't be a problem. Do you think there is a potential security risk here?

src/lib/notificationHandlerWebsocket.js
gip: Long term -- or even soon -- we may want to have a `notificationHandlerAMQP.js` for robust notifications. The config would tell which notification backend is to be used, hence the name.

test/putTransferSpec.js
emschwartz: Why is that test under `describe('GET /accounts/:id/transfers (websocket)', function () {`?

src/services/notificationHandler.js
emschwartz: That's a fair reason for changing it from worker. I'm not sure about handler though. In js event "handlers" are listeners..

src/models/db/notification.js
emschwartz: Doesn't the notification still need to be saved to the DB in order for it to look it up later?

test/putTransferSpec.js
gip: Because the websocket endpoint is `GET /accounts/:id/transfers`
Websocket tests were already at that location before this PR so I just added more there.

src/models/notifications.js
gip: That definitely could be implemented. Do you have a clear use case for that feature at the moment?

src/lib/notificationWorker.js
emschwartz: @justmoon can you weigh in on this, since you implemented the signing in the first place?

src/models/db/notification.js
gip: Any future lookup and can basically be done on `transfers`. We send build the notification from the transfer object.

emschwartz: Oh I see

src/models/notifications.js
emschwartz: If we're removing the notion of notification ids and you would just need to request a transfer state that would probably be sufficient

src/models/transfers.js
gip: We already show a message in the log. Do you suggest that the ledger sends a specific error message through websocket to the client if a transfer isn't found?

test/putTransferSpec.js
emschwartz: Hmm.

Just noticed -- It looks like there is no test for sending notifications with fulfillments attached

src/lib/notificationHandlerWebsocket.js
emschwartz: Wouldn't most of the actual logic be the same? I wonder if we actually would want multiple notification sending classes, versus something like having one that accepts some stream that could either be pure websockets, MQTT or something else

src/models/transfers.js
MatthewPhinney: Not sure that's necessary, but that's one option. To be clear, we are doing it via ws for performance, right? Otherwise, RC could check for transfers that have not been finalized and use the `get transfer` endpoint to get the state of those transfers from the ledger.

src/controllers/accounts.js
```diff
@@ -175,6 +177,13 @@ function * subscribeTransfers () {
   })
 
   this.websocket.on('close', close)
+  // The client can request a notification for a given transfer id
+  this.websocket.on('message', (messageString) => {
+    const message = JSON.parse(messageString)
+    if (message.type && message.type === 'request_notification') {
```
MatthewPhinney: Is this API documented anywhere?

src/lib/notificationHandlerWebsocket.js
```diff
@@ -0,0 +1,52 @@
...
+
+    // Prepare notification for websocket subscribers
+    const notificationBody = {
+      resource: transfer.getDataExternal()
+    }
+
+    // If the transfer is finalized, see if it was finalized by a fulfillment
+    let fulfillment
+    if (transfer.isFinalized()) {
+      fulfillment = yield this.Fulfillment.findByTransfer(transfer.id, { transaction })
```
MatthewPhinney: Are you working off an old copy of the ledger? I don't see `findByTransfer` anywhere in the code.

MatthewPhinney: Separately, how important is it to include the fulfillment in the notification? The client mainly cares about the state of the transfer, right? If they want additional info, they could use the `get transfer` endpoint.

src/services/notificationHandler.js
gip: I was using the more general English meaning. What new name do you suggest? `notificationSender` ?

sentientwaffle: What about `notificationBroadcaster`?

src/lib/notificationHandlerWebsocket.js
gip: Logic would have to be a bit different actually. For robust notifications we would have to start a transaction on `transfers`, then post the message in the amqp queue and then close the transaction when successful. Websocket stuff is really light in comparison.

src/models/db/notification.js
clark800: This already doesn't exist on master, this pull request is based on a very old version of the code...

src/models/transfers.js
gip: That could be a good feature. Let's do that when we have a clear use case as why this would be required. That will be easy to add later.

src/lib/notificationHandlerWebsocket.js
gip: Good point. The original notification logic was including that in the notification so I just left it. There may be a use case for that (beyond debugging).

test/putTransferSpec.js
gip: Do you think we should add a test to check the fulfillment's values?

clark800:
"Breaking" and "Refactor" are incompatible: "Code refactoring is the process of restructuring existing computer code—changing the factoring—without changing its external behavior." https://en.wikipedia.org/wiki/Code_refactoring

gip:
A new PR has been created with exactly the same feature: https://github.com/interledger/five-bells-ledger/pull/279

Closing this one. All comments will be addressed. Sorry for the confusion of having a second PR.

emschwartz:
@gip why did you create a new PR? If it was just for renaming, for the future you can rename pull requests without closing them and opening a new one

PR #279, opened 2016-06-29
Breaking: Full support for websocket-based notifications and eol'ed http notifications

gip:
* Removed http notifications and associated db tables / logic
* Client is able to ask for notification being resend
* Associated PR: https://github.com/interledger/five-bells-receiver/pull/3


src/controllers/accounts.js
```diff
@@ -177,6 +179,13 @@ function * subscribeTransfers () {
   })
 
   this.websocket.on('close', close)
+  // The client can request a notification for a given transfer id
```
gip: @MatthewPhinney I'm not ware of us having any documentation for the websocket protocol as of yet. Let me know if you happen to know.

src/lib/notificationBroadcasterWebsocket.js
```diff
@@ -0,0 +1,59 @@
...
+  .convertToExternalFulfillment
+
+class NotificationBroadcaster extends EventEmitter {
+  constructor (log, config) {
+    super()
+    this.log = log
+    this.config = config
+  }
+
+  * sendNotifications (transfer, transaction) {
```
gip: @justmoon To follow up -- the updated implementation doesn't sign its notification. The idea is that the authentication is done when the websocket connection is established. Sounds good?

src/controllers/notifications.js
```diff
@@ -1,67 +0,0 @@
-'use strict'
```
gip: From @emschwartz : Is there a way to add the apidoc comments back but for the websocket connection?

The apidoc comments for the websocket connection are available at https://github.com/interledger/five-bells-ledger/blob/ws_notifications/src/controllers/accounts.js#L126

test/fulfillmentSpec.js
```diff
@@ -332,162 +326,4 @@ describe('PUT /fulfillment', function () {
     expect((yield getAccount('alice')).balance).to.equal(100)
     expect((yield getAccount('bob')).balance).to.equal(0)
   })
-
-  it('should trigger subscriptions when notification is executed', function * () {
```
gip: Tests for notification when transfers are executed / rejected can be found in https://github.com/interledger/five-bells-ledger/blob/ws_notifications/test/accountSpec.js#L771 and https://github.com/interledger/five-bells-ledger/blob/ws_notifications/test/accountSpec.js#L825

Not sure why they were put in `accountSpec.js` in the first place.

src/lib/notificationBroadcasterWebsocket.js
```diff
@@ -0,0 +1,59 @@
...
+      .flatten().pluck('account').value()
+    affectedAccounts.push('*')
+
+    // Prepare notification for websocket subscribers
+    const notificationBody = {
+      resource: convertToExternalTransfer(transfer)
+    }
+
+    // If the transfer is finalized, see if it was finalized by a fulfillment
+    let fulfillment
```
MatthewPhinney: As I mentioned on #277, unless we have a good reason for including the condition fulfillment in the notification, I'd advocate (in the interests of performance) removing this extra DB read. The client may fetch the fulfillment on an as-needed basis via the `GET /transfers/:id/fulfillment` endpoint.

src/lib/notificationBroadcasterWebsocket.js
```diff
@@ -0,0 +1,59 @@
...
+  .convertToExternalTransfer
+const getFulfillment = require('../models/db/fulfillments').getFulfillment
+const convertToExternalFulfillment = require('../models/converters/fulfillments')
+  .convertToExternalFulfillment
+
+class NotificationBroadcaster extends EventEmitter {
+  constructor (log, config) {
+    super()
+    this.log = log
+    this.config = config
```
MatthewPhinney: Doesn't look like `config` is being used. If it's not needed, we may as well remove it from this file and `services/notificationBroadcaster.js`.

test/accountSpec.js
```diff
@@ -689,6 +686,88 @@ describe('Accounts', function () {
...
+      // TODO: Is there a more elegant way?
+      yield timingHelper.sleep(50)
+
+      const transferId = transfer.id.match('http://localhost/transfers/(.+)')[1]
+      this.socket.send(JSON.stringify({ type: 'request_notification',
+                                        id: transferId }))
+
+      yield timingHelper.sleep(50)
+
+      sinon.assert.calledTwice(listener)
```
MatthewPhinney: This is called twice because the first is when the transfer is put, and the second is because we explicitly request a notification via the ws connection? If that's the case, it seems like this could be two tests. 1. Check notification is sent when transfer is PUT to ledger. 2. Check notification is sent when we request one for a transfer already in the db.

src/lib/notificationBroadcasterWebsocket.js
emschwartz: For Universal it is essential that the fulfillment be included with the notification. The connector needs to get that as quickly as possible to pass it on and doing a `GET` would make it much slower

test/fulfillmentSpec.js
emschwartz: Line 395 of this file tests that the fulfillment was attached with the notification. Could you add a test in the `accountSpec.js` that tests for that as well? Also, if you don't think it makes sense to have those tests in that file (I'd definitely agree), you could move them to a more sensible place

test/accountSpec.js
gip: For (2), how would you make sure a transfer is already in the db? If it is by doing a PUT then it looks like (1) would be a subset of (2) and therefore isn't necessary?

src/lib/notificationBroadcasterWebsocket.js
MatthewPhinney: That's a good reason.

test/accountSpec.js
MatthewPhinney: There is an [addTransfers](https://github.com/interledger/five-bells-ledger/blob/master/test/helpers/db.js#L42) helper method. I think it's clearer to have two tests, for the separation of logic, but this test encompasses both cases, so I don't think this issue needs to hold up the PR.

test/fulfillmentSpec.js
MatthewPhinney: Yeah, I'd also argue for removing those tests from `accountSpec`.

gip: Done: moved relevant tests to `notificationSpec.js` and added new tests to check the fulfillment. See https://github.com/interledger/five-bells-ledger/blob/ws_notifications/test/notificationSpec.js#L344

src/models/transfers.js
```diff
@@ -481,11 +473,27 @@ function * insertTransfers (externalTransfers) {
     converters.convertToInternalTransfer))
 }
 
+function * resendNotification (transferId) {
```
MatthewPhinney: I think you want some kind of check here to ensure the party requesting the notification has privileges to views this transfer ... although it doesn't look like the get transfer endpoint has any sort of authentication either.

gip: Good points, but actually I don't think it's a problem because:
* When a request is send, a lookup is made in the database. If the transfer exists, the notification will first be emitted using the event name `transfer-{$ACCOUNT}`. See https://github.com/interledger/five-bells-ledger/blob/ws_notifications/src/lib/notificationBroadcasterWebsocket.js#L52
* So the notification will only be sent to the ws connections that listen to that account, no worries there

That being said, if a client sends a request with a transfer id which correspond to a transfer that doesn't belong to that account:
* A notification will be sent to the clients listening for notifications on that account, even though they have not requested a new notification
* I don't think that is a problem as this is very unlikely to happen. The probability of collision for generated UUIDs is probably lower than the risk of being reincarnated as an olive in a next life

There is an issue with this implementation though. If 2 (or more) clients are listening for notifications for a given account and only one of the client requests a new notification for a given transfer. Then _all_ the clients will receive the notification, not only the requesting one. We could either:
* Change this behaviour (that would take some work)
* Add to the specs that notification may be sent more than once. Clients would therefore need to be idempotent

MatthewPhinney: Yeah, that makes sense. I think we can expect clients to be robust enough to handle duplicate notifications, should they occur.

src/controllers/notifications.js
emschwartz: It doesn't look like there's a standard way to do apidoc for websockets (just submitted a [question](https://github.com/apidoc/apidoc/issues/501) on the apidoc repo). Not sure what we should do for now but we should definitely document this API...

gip: It is kinda documented in the link I added above (https://github.com/interledger/five-bells-ledger/blob/ws_notifications/src/controllers/accounts.js#L126). Isn't it sufficient for now?

emschwartz: Maybe, but for example the feature you're adding now (requesting a notification resend) is not documented anywhere

gip: Updated the apidoc portion of the file. We also should consider writing additional docs outlining the different guarantees that apply to notifications from the ledger and explain how a client could implement robust notifications on top of the ledger if needed by an application. (I would be in favor of open sourcing a client implementing robust notifications actually).

gip: It is already possible to implement pseudo-robust notifications with the current ledger using some retry logic if a notification is missed.

emschwartz: How does a client know if they missed a notification?

gip: By checking the expiry: resending a new notification request after a transaction should have expired.

MatthewPhinney: If the ws connection disconnects/reconnects, or the client has a transfer that is past expiry, they can query the ledger to get the transfer state.

emschwartz: That's assuming we're talking about the client who sent the transfer originally, and that that client has state. If you're listening for incoming transfers you'll have no idea you missed something if you never heard about it in the first place. We also shouldn't assume that clients or connectors are keeping state. Many will rely on the ledger for state.

gip: Yes, we're talking about a client with some database-backed state (that's why I called it _pseudo-robust).

emschwartz: Ok, well in any case I think we need to implement reliable notifications on the ledger's side

justmoon:
Agree on taking out the REST hooks. This will significantly improve ledger performance.

Not sure why the resend functionality is being added. Instead of requesting a resend, what's wrong with simply `GET`ting the transfer (and fulfillment if needed.) That seems overall more efficient for the client because they get the transfer in the response instead of getting it separately as a notification, which would mean that they have to reconstruct all the context again such as looking up the transfer from the database etc.

I'm assuming that this design was chosen in order to avoid duplicating the code that currently exists in the websocket notification handler. May I suggest simply refactoring the app such that the notification handler is a reusable function which can be called both from the notification handler and when manually requesting the transfer from the ledger. This will be simpler, more performant and removes the need for the resend feature in the ledger.

gip:
The resend notification feature may be better in term if performance and stability for the ledger - doing things asynchronously usually brings better performance in large distributed systems. The ledger would have better control over what it being executed. Plus there would be a unique channel for a client to get notifications.

I don't think this (huge) PR should be delayed by this discussion however. In the interest of speed, I have removed the resend logic for now. I'll do a new PR once we've discussed further and find an agreement. Can you please review @justmoon.

justmoon:
LGTM

Improves performance and I'm happy to see this much dead code removed. :+1:

PR #281, opened 2016-07-01
Use integer for primary key in L_TRANSFERS

clark800:


src/models/db/transfers.js
```diff
@@ -69,42 +71,64 @@ function convertToPersistent (data) {
...
 }
 
 function * updateTransfer (transfer, options) {
-  return db.update(transfer, {TRANSFER_ID: transfer.id},
-    options && options.transaction).then((result) => {
-      return adjustments.upsertAdjustments(transfer, options).then(() => result)
+  const transaction = options && options.transaction
+  return db.update(transfer, {TRANSFER_UUID: transfer.id}, transaction)
+  .then((result) => {
+    return db.selectOne({TRANSFER_UUID: transfer.id}, transaction)
```
MatthewPhinney: So the issue here is that knex won't return the autogenerated ID on an update?

clark800: Yeah, `update` "Resolves the promise / fulfills the callback with the number of affected rows for the query."

src/models/db/transfers.js
```diff
@@ -69,42 +71,64 @@ function convertToPersistent (data) {
...
     .then((transfer) => {
-      return adjustments.getAdjustments(id, options).then((adjustments) => {
+      if (transfer === null) {
+        return null
+      }
+      return adjustments.getAdjustments(transfer._id, options)
+      .then((adjustments) => {
         const result = _.assign({}, transfer, adjustments)
-        return _.isEmpty(result) ? null : result
+        return _.isEmpty(result) ? null : _.omit(result, '_id')
```
MatthewPhinney: Might be clearer to omit the `_id` in the `convertToExternal`.

MatthewPhinney:
LGTM

src/models/db/transfers.js
clark800: Then we wouldn't be able to use `selectOne`, which creates a lot more duplicated code

Issue #276, opened 2016-06-28
Support reliable notifications


justmoon:
We are about to remove reliable notification support via REST hooks. Do we need a new facility (e.g. using MQTT/[mosca](https://github.com/mcollina/mosca)) to replace it?

gip:
AMQP could also be an option if very reliable notifications are needed.

PR #285, opened 2016-07-08
Add endpoint to fetch transfer by execution condition

mwshortt:
This allows a client to not need to keep track of the transfer id if they know the execution condition


mwshortt:
This PR would alleviate the need for the more hacky fix of https://github.com/interledger/five-bells-connector/pull/184 @emschwartz @justmoon

test/data/transfers/withExpiry.json
```diff
@@ -17,7 +17,7 @@
     "account": "http://localhost/accounts/dave",
     "amount": "40"
   }],
-  "execution_condition": "cc:0:3:vmvf6B7EpFalN6RGDx9F4f4z0wtOIgsIdCmbgv06ceI:7",
+  "execution_condition": "cc:0:3:Xk14jPQJs7vrbEZ9FkeZPGBr0YqVzkpOYjFp_tIZMgs:7",
```
MatthewPhinney: What was wrong with the other condition?

src/models/db/transfers.js
```diff
@@ -90,6 +90,20 @@ function * getTransfer (uuid, options) {
     })
 }
 
+function * getTransferByExecutionCondition (executionCondition, options) {
+  return db.selectOne({EXECUTION_CONDITION: executionCondition}, options && options.transaction)
```
MatthewPhinney: If you're going to query the db by execution condition, please add an index to the `EXECUTION_CONDITION` column in `src/sql/pg/create.sql` and `src/sql/strong-oracle/create.sql` (we don't really care about performance on sqlite), as done for `TRANSFER_UUID`.

src/models/db/transfers.js
```diff
@@ -90,6 +90,20 @@ function * getTransfer (uuid, options) {
     })
 }
 
+function * getTransferByExecutionCondition (executionCondition, options) {
```
clark800: Looks like a good case for extracting a method to reduce duplication between this method and `getTransfer`

clark800:
> This allows a client to not need to keep track of the transfer id if they know the execution condition

Can you explain why this is necessary?

justmoon:
I'll have to wait until I get back to the hotel to review properly, but this is definitely a better solution than interledger/five-bells-connector#184 because this will work even for multi hop paths with mutually distrustful parties. Good thinking! :+1:

emschwartz:
This is definitely a better solution than https://github.com/interledger/five-bells-connector/pull/184.

The only potential issue I could see with this is that it would mean you cannot submit two transfers with the same condition. I know we said that it would be an error to submit a transfer with a condition that had already been fulfilled. However, I could still see a use case for submitting multiple transfers with the same condition. Basically, if you want to send multiple small payments through different paths to a single recipient and have them all execute upon the receipt of a single fulfillment. @justmoon thoughts?

test/data/transfers/withExpiry.json
mwshortt: Nothing was wrong, they all had the same execution conditions and I needed one that was different (so I could get a 404 in the unit test)

src/models/db/transfers.js
mwshortt: Will do! Thanks for pointing that out!

src/models/db/transfers.js
mwshortt: Sounds good! I'll add that method

clark800: I actually already did this in my pull request, so you could copy from there: https://github.com/interledger/five-bells-ledger/pull/286/files#diff-fb6e03a84a05cfd77df4bcbfa0e28b33R92

mwshortt:
@clark800 Sure! So with the new `ilp_header` format we are no longer able to specify the transfer id for the destination transfer. Since the websocket notifications are not robust if someone listening to the receiving ledger goes down when a notification is broadcast there is no way for them to check the state of that transfer on the ledger because they don't know the transfer id. If they have the execution condition, which is the use case I am working on, then they would be able to query the ledger using that.

src/models/db/transfers.js
mwshortt: Excellent, ty!

clark800:
Is the `execution_condition` guaranteed to always be unique if it's non-null? If so we could construct the transfer UUID from the hash of the execution condition so that we wouldn't need this endpoint.

emschwartz:
@clark800 that would present the same problem described [here](https://github.com/interledger/five-bells-connector/blob/2dffa0063d48e58ddd9b154ac37d45f334caa1fd/src/lib/route-builder.js#L131-L141)

clark800:
@emschwartz Good point. Could that be addressed by salting the crypto-conditions?

emschwartz:
Then the receiver wouldn't be able to know beforehand what the id will be -- I think that would defeat the point. That id issue really means that no one aside from the connector should know what the id will be beforehand.

I just realized that doing this based on the crypto conditions -- while enforcing that crypto conditions can only be used for one transfer -- actually presents the same problem. I can create another transfer with the same crypto condition and different details and mess with the connector in the same way. This suggests that we **cannot** enforce uniqueness on crypto conditions. This `GET` API for transfers by their conditions might need to return an array.

MatthewPhinney:
@emschwartz yeah, I think this endpoint should return all transfers executed by the fulfillment.

emschwartz:
@MatthewPhinney you mean all transfers with that condition (not fulfillment), right?

MatthewPhinney:
Yeah, that's what I meant.

clark800:
@emschwartz @justmoon What do you think of putting the `payment_id` in the `additional_info` field and looking up based on that? It might be a bit more work, but it seems less hacky and should make the lookup unique.

emschwartz:
It seems hackier to me to create an index based on something in a field
called additional_info. I think it seems sensible to be able to look up
transfers by their conditions

clark800:
>  I think it seems sensible to be able to look up transfers by their conditions

Without knowing the use case in advance, it took me a while to understand why someone would want to do this. After Madeline explained it I realized the problem we're trying to solve is to find a way to identify transfers by some information that is available prior to the transfer ID creation. The `execution_condition` happens to satisfy this property, but it is not fundamentally designed for that purpose, which means we would be effectively repurposing something because it was conveniently available. The fact that the `execution_condition` is not guaranteed to be unique is evidence that it isn't really ideal to serve as an identifier. Regarding an index on a field in `additional_info`, this may be a general enough problem to justify making an optional top-level field in the transfer object.

emschwartz:
@clark800 that's not entirely true. We've described crypto conditions at various points as a system for handling events in distributed systems. The question is whether multiple "events" (on the same system) should be able to be triggered by the same condition fulfillment.

> Regarding an index on a field in additional_info, this may be a general enough problem to justify making an optional top-level field in the transfer object.

This sounds like a receiver-specified index. I think this sort of thing would be better put in the credit memo. Don't love the idea of having a lookup on the memo fields though.

Since this discussion was prompted because we removed support for reliable notifications, I think it may be best addressed by reimplementing [reliable notifications using MQTT](https://github.com/interledger/five-bells-ledger/issues/276).

mwshortt:
I've updated this PR to support returning an array of transfers from this endpoint.
@emschwartz @clark800 while supporting reliable websocket notifications may be the best solution in the long term, we need a solution in the interim and I think this endpoint is a good solution for right now. 
Let me know if you agree and we can get this reviewed and merged!

justmoon:
Looks good, only thing I would change is the URL - `/transfers/*` should be reserved for looking up a transfer by ID. I can think of three options:

1. `/conditions/:condition/executingTransfers` or `/conditions/:condition/transfersExecutedBy` - I.e. we imagine a condition resource and we want to get all of the transfers executed by that resource.

2. `/transfers?execution_condition=` - I.e. we want to query the collection of all transfers by execution condition.

3. `/transfers/byExecutionCondition/:condition` - I.e. we are talking about the transfers collection, but we have a more specific endpoint for querying by condition.

Option 2 would be better only if we're planning to expand this into a general "transfer search" function. I'd go with option 3, because it's easy to read it and understand what it does.

test/getTransferSpec.js
```diff
@@ -132,8 +133,28 @@ describe('GET /transfers/:execution_condition/execution_condition', function ()
...
+
+  it('should return 200 for multiple existing transfers', function * () {
+    yield dbHelper.addTransfers([this.preparedTransfer])
+
+    const transfer = this.executedTransfer
+    yield this.request()
+      .get(`/transfers/${transfer.execution_condition}/execution_condition`)
+      .expect(200)
+      .expect((res) => {
+        const sortedResponse = _.sortBy(res.body, (transfer) => {
```
MatthewPhinney: Why do you need to sort the response? Won't the endoint return the the transfers in the same order each time?

MatthewPhinney:
I think this is good for now. In the long term we'll need to support reliable notifications, in which case we won't need this endpoint.

test/getTransferSpec.js
mwshortt: Not necessarily. It depends on the order they are returned from the database since we aren't doing any sorting.

mwshortt:
I agree with Option 3, `/transfers/byExecutionCondition/:condition`. I've updated the PR, let me know if I'm good to merge it!

emschwartz:
LGTM. @justmoon?

justmoon:
LGTM!

PR #288, opened 2016-07-12
Fix /accounts/*/transfers subscription. Fixes #287

### LINKED ISSUE #287: Can't subscribe to /accounts/*/transfers
Schema validation fails [here](https://github.com/interledger/five-bells-ledger/blob/8cfe503ea8d1489ef21a78be18f61f4cc746ce73/src/controllers/accounts.js#L176) on a  `*`.

vhpoet:


MatthewPhinney:
What error does this throw if a non-admin user tries to subscribe to *? Is there a test for this case?

vhpoet:
@MatthewPhinney fails the validation

MatthewPhinney:
LGTM

gip:
At the minimum, I'd think we need a test to make sure if a client subscribe to `*` and is not admin there is an error.

PR #295, opened 2016-07-19
Add DB audit fields and remove unused code

clark800:


clark800:
@lumberj It looks like the docker image doesn't contain the `psql` command. I got the tests passing by not using docker for the postgres tests. Does this sound like a good approach?

src/lib/db.js
```diff
@@ -50,9 +50,33 @@ function executeSQLPlus (sqlFilepath) {
...
+      '--port=' + (connection.port || 5432),
+      '--dbname=' + connection.database,
+      '--file=' + path.resolve(sqlFilepath)
+    ]
+    const env = {
+      PATH: process.env.PATH,
+      PGPASSWORD: connection.password
+    }
+    const childProcess = spawn(command, args, {env})
+    childProcess.on('close', (code) => {
```
lumberj: Does there need to be an `on('error')` handler?

lumberj:
LGTM

src/lib/db.js
clark800: Yeah, I'll add that

scripts/ci-test.sh
```diff
@@ -29,7 +29,7 @@ mysqltest() {
 postgrestest() {
   psql -U ubuntu -c 'DROP DATABASE circle_test;'
   psql -U ubuntu -c 'CREATE DATABASE circle_test;'
-  docker run --name=ledger-test-postgres -it --net=host -e LEDGER_UNIT_DB_URI=postgres://ubuntu@localhost/circle_test interledger/five-bells-ledger npm test
+  LEDGER_UNIT_DB_URI=postgres://ubuntu@localhost/circle_test npm test
```
MatthewPhinney: Why are we not using docker for this?

src/lib/db.js
```diff
@@ -64,6 +90,8 @@ function executeScript (filename) {
 
   if (dbType === 'strong-oracle') {
     return executeSQLPlus(filepath)
+  } else if (dbType === 'pg') {
+    return executePSQL(filepath)
```
MatthewPhinney: So now `executeStatements` is only used for sqlite? What's the advantage of doing it this way? Is this working around a `knex` issue?

scripts/ci-test.sh
clark800: The docker image doesn't include the `psql` binary, which is needed

src/lib/db.js
clark800: There are scripts in the SQL file that don't work when run line by line: https://github.com/interledger/five-bells-ledger/blob/master/src/sql/pg/create.sql#L191

PR #299, opened 2016-07-26
Add db health check (RILP-508)

MatthewPhinney:


src/lib/db.js
```diff
@@ -128,6 +128,19 @@ function * truncateTables () {
...
 
+function * isConnected () {
+  const query = knex.client.config.client === 'strong-oracle'
+    ? 'SELECT 1 FROM DUAL' : 'SELECT 1'
+  return knex.raw(query)
+  .then(() => {
+    return true
+  })
+  .catch((e) => {
+    console.log('DB error: ', e)
```
clark800: I think you have to use bunyan format logging: https://github.com/interledger/five-bells-ledger/blob/master/package.json#L15

MatthewPhinney: Oh yeah ... I'm going to remove this line anyway. I only added it because I was having trouble debugging locally.

clark800:
LGTM

PR #306, opened 2016-08-03
Remove oracle public synonyms to allow multiple ledger instances to r…

clark800:
…un on a single database instance under different usernames


gip:
LGTM once tests pass

PR #307, opened 2016-08-04
Add mssql support

clark800:


clark800:
Note: We will no longer be able to make pull requests from forks after this is merged because we get the MSSQL credentials from CircleCI environment variables and they are not exported for fork builds for security of secrets.

package.json
```diff
@@ -17,6 +17,7 @@
     "lint": "eslint .",
     "test": "istanbul test -- _mocha",
     "sqlplus": "LD_LIBRARY_PATH=/opt/oracle/instantclient DYLD_LIBRARY_PATH=/opt/oracle/instantclient /opt/oracle/instantclient/sqlplus system/oracle@localhost:1521/xe",
+    "test-mssql": "LEDGER_UNIT_DB_URI=mssql://$MSSQL_USERNAME:$MSSQL_PASSWORD@in-1808.cojrajw6pfyj.us-west-2.rds.amazonaws.com:1433/$MSSQL_DATABASE node node_modules/.bin/istanbul test -- _mocha",
```
MatthewPhinney: We are running this on AWS because there is no docker image? I missed the discussion on why this is different than our setup for Oracle.

package.json
```diff
@@ -42,7 +43,7 @@
     "five-bells-condition": "^3.2.0",
     "five-bells-shared": "^17.0.0",
     "keypair": "^1.0.0",
-    "knex": "^0.10.0",
+    "knex": "^0.11.9",
```
MatthewPhinney: What features of knex 11 do you need?

scripts/ci-test.sh
```diff
@@ -20,6 +20,24 @@ dockerBuild() {
   docker build -t interledger/five-bells-ledger .
 }
 
+mssqltest() {
+  : ${MSSQL_USERNAME?"MSSQL_USERNAME not set; note that CircleCI does not export environment variables on forks for security of secrets"}
+  MSSQL_DATABASE="CIRCLECI_$(openssl rand -hex 6)"
```
MatthewPhinney: Is the idea to have a different db each time? Is there any issue with using the same name each time?

package.json
clark800: MSSQL only runs on Windows and Docker only runs linux

package.json
clark800: The older version used a different name for the hostname in the config for mssql only

src/lib/holds.js
```diff
@@ -26,31 +26,34 @@ function insertEntryByName (accountName, transferId, transaction) {
 
 function holdFunds (transfer, transaction) {
   return Promise.all(transfer.debits.map((debit) => {
-    return Promise.all([
-      adjustBalance(debit.account, -debit.amount, transaction),
-      adjustBalance('hold', debit.amount, transaction),
-      insertEntryByName(debit.account, transfer.id, transaction)
-    ])
+    return adjustBalance(debit.account, -debit.amount, transaction).then(() => {
```
MatthewPhinney: Was this not done in sequence before?

scripts/ci-test.sh
clark800: If you use the same name, then you wouldn't be able to do concurrent tests

src/lib/holds.js
clark800: No, see the diff, `Promise.all` does them in "parallel" (quotes because javascript is single-threaded)

src/sql/mssql/create.sql
```diff
@@ -0,0 +1,253 @@
...
+
+ALTER TABLE [L_ENTRIES] ADD CONSTRAINT FK_TRANSFER_ID__ENTRIES FOREIGN KEY
+  ([TRANSFER_ID]) REFERENCES [L_TRANSFERS] ([TRANSFER_ID])
+GO
+
+
+CREATE TABLE [L_FULFILLMENTS]
+(
+  [FULFILLMENT_ID] INT NOT NULL CONSTRAINT L_PK_FULFILLMENTS PRIMARY KEY,
+  [TRANSFER_ID] INT NOT NULL,
```
MatthewPhinney: Do you need to index this column?

clark800: Yeah, I think it should be

MatthewPhinney:
I think you need to update here too: https://github.com/interledger/five-bells-ledger/blob/3e46a7b6d755182ca57a19f8be7cc81f3a965019/src/models/transfers.js#L337-L341. I wonder why the `assert` doesn't throw.

package.json
MatthewPhinney: Oh yeah, Microsoft ...

Issue #308, opened 2016-08-05
Provide an ILP address prefix in the ledger metadata


justmoon:
We should add a config option `LEDGER_ILP_PREFIX=ilpdemo.red.` which will be added to the ledger metadata as follows:

```js
{
  "currency_code": "USD",
  "currency_symbol": "$",
  //...
  "ilp_prefix": "ilpdemo.red."
}

justmoon:
Implemented in #309.

PR #309, opened 2016-08-05
feat: add config option for ILP prefix

bensharaf:
Addresses https://github.com/interledger/five-bells-ledger/issues/308, and required in order to address https://github.com/interledger/js-ilp-plugin-bells/issues/21

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ npm run docs
 
 Open `apidocs-out/index.html` in a web browser to see the generated API documentation.
 
-
 ## Running with Docker (Alternative Method)
 
 This project can be run in a [Docker](https://www.docker.com/) container.
--- a/src/controllers/metadata.js
+++ b/src/controllers/metadata.js
@@ -1,26 +1,4 @@
 'use strict'
-const config = require('../services/config')
-
-const base = config.getIn(['server', 'base_uri'])
-const metadata = {
-  currency_code: config.getIn(['currency', 'code']),
-  currency_symbol: config.getIn(['currency', 'symbol']),
-  condition_sign_public_key: config.getIn(['keys', 'ed25519', 'public']),
-  notification_sign_public_key: config.getIn(['keys', 'notification_sign', 'public']),
-  urls: {
-    health: base + '/health',
-    transfer: base + '/transfers/:id',
-    transfer_fulfillment: base + '/transfers/:id/fulfillment',
-    transfer_state: base + '/transfers/:id/state',
-    connectors: base + '/connectors',
-    accounts: base + '/accounts',
-    account: base + '/accounts/:name',
-    subscription: base + '/subscriptions/:id',
-    subscription_notification: base + '/subscriptions/:subscription_id/notifications/:notification_id'
-  },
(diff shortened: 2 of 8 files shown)


emschwartz:
Please add to the [config](https://github.com/interledger/five-bells-ledger/blob/master/test/configSpec.js) and [metadata](https://github.com/interledger/five-bells-ledger/blob/master/test/metadataSpec.js) tests.

bensharaf:
Added the field to the metadata spec and config sepc

test/metadataSpec.js
```diff
@@ -26,6 +26,7 @@ describe('Metadata', function () {
           expect(res.body).to.deep.equal({
             currency_code: null,
             currency_symbol: null,
+            ilp_prefix: null,
```
emschwartz: It would be better to test this such that the value isn't null

bensharaf: in configSpec https://github.com/interledger/five-bells-ledger/pull/309/files#diff-2ed4457106d98f881aab4dcbdccd6b8dR81 it tests setting the value

emschwartz: Yes but it doesn't test to make sure that when you set it it actually shows up in the metadata. Testing if a value is null in this case doesn't actually test anything

bensharaf: due to some weirdness in the config class, I'm blocked by https://github.com/interledger/five-bells-ledger/pull/310 to include the test I wrote

bensharaf: fixed now

README.md
```diff
@@ -70,6 +70,7 @@ With the following configuration options set as environment variables:
 * `LEDGER_FEATURE_CREDIT_AUTH` (default: `0`) whether or not to require credits to be authorized.
 * `LEDGER_CURRENCY_CODE` (default: none) ISO 4217 currency code
 * `LEDGER_CURRENCY_SYMBOL` (default: none) currency symbol
+* `LEDGER_ILP_PREFIX` (default: none) ledger ILP addresss
```
justmoon: Typo: Should be "address"

src/controllers/metadata.js
```diff
@@ -32,4 +10,29 @@ const metadata = {
  *
  * @returns {void}
  */
-exports.getResource = function * () { this.body = metadata }
+module.exports = (config) => ({
+  getResource: function * () {
+    const base = config.getIn(['server', 'base_uri'])
+    const metadata = {
```
justmoon: What's the rationale for re-generating this object on every request? Since the config is immutable, you should generate it once when the outer function is called and then just return it. Probably not a huge difference performance-wise, but it also makes the code more readable imho.

bensharaf: The problem was that before, metadata would only be re-generated once (when it was imported the first time). I can make it generate once each time that the module.exports is called.

justmoon:
LGTM if tests pass

bensharaf:
blocked by this https://github.com/interledger/five-bells-ledger/pull/315#issuecomment-240561233

mDuo13:
@bensharaf Thanks for fixing that for me.

PR #310, opened 2016-08-05
fix: use the app's config when creating metadata

bensharaf:
Right now `src/controllers/metadata` includes `src/services/config`. If a different config is passed into the `modules` argument of an App, then the metadata will not match the App's config.


emschwartz:
If this is a fix you should include tests to make sure the same bug isn't re-introduced later

bensharaf:
Moved fix to https://github.com/interledger/five-bells-ledger/pull/309

PR #313, opened 2016-08-08
fix: handle no username DB_URI

vhpoet:
The ledger would throw an exception with
```
LEDGER_SYNC_TRUE=1
LEDGER_DB_URI=postgres://localhost/wallet1-ledger
```


justmoon:
LGTM

PR #319, opened 2016-08-29
feat: getFulfillment {Transfer,Fulfillment}NotFoundError

sentientwaffle:
Related to https://github.com/interledger/js-ilp-plugin-bells/pull/29 and https://github.com/interledger/rfcs/pull/84#issuecomment-242001174

### DIFF
--- a/src/models/db/fulfillments.js
+++ b/src/models/db/fulfillments.js
@@ -5,6 +5,8 @@ const db = require('./utils')('L_FULFILLMENTS',
   convertToPersistent, convertFromPersistent)
 const getTransferId = require('./transfers').getTransferId
 const removeAuditFields = require('./audit').removeAuditFields
+const TransferNotFoundError = require('../../errors/transfer-not-found-error')
+const FulfillmentNotFoundError = require('../../errors/fulfillment-not-found-error')
 
 function convertFromPersistent (data) {
   const result = _.mapKeys(_.cloneDeep(data), (value, key) => key.toLowerCase())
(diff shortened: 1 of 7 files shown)


src/models/db/fulfillments.js
```diff
@@ -39,23 +41,36 @@ function * insertFulfillments (fulfillments, options) {
...
   return getTransferId(transferUuid, options).then((transferId) => {
+    if (!transferId) {
+      throw new TransferNotFoundError('This transfer does not exist')
+    }
     return db.selectOne({TRANSFER_ID: transferId},
       options && options.transaction).then((result) => {
-        if (result) {
-          result.transfer_id = transferUuid
+        if (!result) {
+          throw new FulfillmentNotFoundError('This transfer has no fulfillment')
```
emschwartz: It might be good to differentiate at least in the error messages between a transfer that is not yet finalized (`This transfer has not yet been fulfilled`) and one that has already expired (`This transfer expired before it was fulfilled`)

src/models/db/fulfillments.js
```diff
@@ -38,15 +42,30 @@ function * insertFulfillments (fulfillments, options) {
...
+  return db.selectOne({TRANSFER_ID: transferId},
+    options && options.transaction).then((result) => {
+      if (!result) {
+        if (transfer.expires_at && moment().isAfter(transfer.expires_at)) {
+          throw new FulfillmentNotFoundError('This transfer expired before it was fulfilled')
         }
-        return result
-      })
-  })
+        throw new FulfillmentNotFoundError('This transfer has no fulfillment')
```
emschwartz: Sorry for nit-picking, but could you change it to say "This transfer has not yet been fulfilled"? The current wording sounds very final, even though it's the case where it's not finalized yet

sentientwaffle: Good point; updated

emschwartz:
LGTM once tests pass

PR #320, opened 2016-08-31
feat: PUT /rejection

sentientwaffle:
Reject credits by setting `rejected=true` on PUT, or by PUTing a message to `/transfers/:id/rejection`.

Depends on https://github.com/interledger/five-bells-shared/pull/154


src/models/transfers.js
```diff
@@ -290,6 +302,23 @@ function * processImmediateExecution (transfer, transaction) {
   }
 }
 
+function * processCreditRejection (transfer, transaction) {
+  if (transfer.state === transferStates.TRANSFER_STATE_REJECTED) return
+  const allCreditsAreRejected = _.every(transfer.credits, 'rejected')
```
emschwartz: A transfer should be rejected if a single credit is rejected, no?

test/putTransferSpec.js
```diff
@@ -1483,6 +1483,93 @@ describe('PUT /transfers/:id', function () {
     })
   })
 
+  describe('rejection', function () {
+    it('should return 403 if authorized:true is set for any credits that are ' +
```
emschwartz: `rejected: true` you mean?

test/putTransferSpec.js
```diff
@@ -1483,6 +1483,93 @@ describe('PUT /transfers/:id', function () {
...
+          rejection_reason: 'cancelled',
+          timeline: {
+            rejected_at: '2015-06-16T00:00:00.000Z',
+            proposed_at: '2015-06-16T00:00:00.000Z'
+          }
+        }))
+        .expect(validator.validateTransfer)
+        .end()
+    })
+
```
emschwartz: I think there should also be a test for rejecting a proposed transfer

test/rejectionSpec.js
```diff
@@ -0,0 +1,195 @@
...
+        ],
+        timeline: {
+          prepared_at: '2015-06-16T00:00:00.000Z',
+          proposed_at: '2015-06-16T00:00:00.000Z',
+          rejected_at: '2015-06-16T00:00:00.000Z'
+        }
+      }))
+  })
+
+  it('rejects the transfer when all credits are rejected', function * () {
```
emschwartz: I think it should reject when a single one is rejected

emschwartz:
LGTM

PR #322, opened 2016-09-08
feat: BREAKING remove /connectors; send them as metadata

### LINKED ISSUE #321: Explicitly define "recommended connectors" in ledger config
Currently, the connectors endpoint returns all registered connectors on the ledger.

This is unsafe. Anyone could create a connector and make it return the cheapest quotes. It's useful to be able to get a "default" set of connectors from the ledger. However that list must be curated, so for now lets add a config variable which allows the ledger operator to specify which connectors should be included in the `/connectors` endpoint. (For backwards compatibility, it's fine if `/connectors` returns all connectors by default for now.)

Proposed config:

``` sh
LEDGER_RECOMMENDED_CONNECTORS=connie,alice
LEDGER_RECOMMENDED_CONNECTORS=* # default
```

sentientwaffle:
Fixes #321

### DIFF
--- a/test/configSpec.js
+++ b/test/configSpec.js
@@ -189,4 +189,23 @@ describe('loadConfig', () => {
       expect(loadConfig).to.throw(/Cannot support precision > 15/)
     })
   })
+
+  describe('config.recommendedConnectors', () => {
+    it('defaults to undefined', () => {
+      const _config = loadConfig()
+      expect(_config.recommendedConnectors).to.equal(undefined)
+    })
+
+    it('is undefined when recommendedConnectors is "*"', () => {
+      process.env.LEDGER_RECOMMENDED_CONNECTORS = '*'
+      const _config = loadConfig()
+      expect(_config.recommendedConnectors).to.equal(undefined)
+    })
+
+    it('is a list of connector names, otherwise', () => {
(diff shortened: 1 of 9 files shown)


test/configSpec.js
```diff
@@ -189,4 +189,23 @@ describe('loadConfig', () => {
...
     })
   })
+
+  describe('config.recommendedConnectors', () => {
+    it('defaults to undefined', () => {
+      const _config = loadConfig()
+      expect(_config.recommendedConnectors).to.equal(undefined)
+    })
+
+    it('is undefined when recommendedConnectors is "*"', () => {
```
emschwartz: Does `*` mean everyone is recommended?

sentientwaffle: All connector accounts. Its the current (and default) behavior.

emschwartz: Got it. What happens if you configure `RECOMMENDED_CONNECTORS` with an account that doesn't have the `connector` field set?

sentientwaffle: Any accounts listed in `RECOMMENDED_CONNECTORS` that aren't actually connector's will be excluded from the returned list. If they are later changed to connectors, they will get added in automatically.

emschwartz:
LGTM once tests pass

PR #325, opened 2016-09-19
feat: send ledger-plugin RFC errors

sentientwaffle:
Depends on https://github.com/interledger/five-bells-shared/pull/156

### DIFF
--- a/test/fulfillmentSpec.js
+++ b/test/fulfillmentSpec.js
@@ -67,8 +67,43 @@ describe('GET /fulfillment', function () {
       .auth('admin', 'admin')
       .expect(404)
       .expect({
-        id: 'FulfillmentNotFoundError',
+        id: 'MissingFulfillmentError',
         message: 'This transfer has not yet been fulfilled'
+      })
+      .end()
+  })
+
+  it('should return AlreadyRolledBackError if the transfer is rejected', function * () {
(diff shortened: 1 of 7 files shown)


test/fulfillmentSpec.js
```diff
@@ -342,9 +409,11 @@ describe('PUT /fulfillment', function () {
...
       .put(transfer.id + '/fulfillment')
       .send(this.cancellationConditionFulfillment)
-      .expect(200)
-      .expect(this.cancellationConditionFulfillment)
-      .expect(validator.validateFulfillment)
+      .expect(422)
+      .expect({
+        id: 'AlreadyRolledBackError',
+        message: 'This transfer has already been rejected'
+      })
```
emschwartz: I think canceling a transfer with a `PUT` should be idempotent. It shouldn't return an error, just acknowledge the request as if it were the first one.

sentientwaffle: Ok. I was going off of [the spec](https://github.com/interledger/rfcs/blob/master/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#fulfillcondition) (see `AlreadyRolledBackError`).

emschwartz: Oh actually my mistake, I misunderstood which function this was testing.

PR #331, opened 2016-09-26
refactor: improve ledger performance. fixes #330

### LINKED ISSUE #330: InsufficientFundsError on hold account
I'm stress testing the wallet by submitting multiple (2-3) parallel payments and sometimes I get this strange exception from the ledger when the plugin-bells tries to submit the fulfillment.

```
{"id":"InsufficientFundsError","message":"Sender has insufficient funds.","owner":"hold"}
```

vhpoet:
See the discussion in https://github.com/interledger/five-bells-ledger/issues/330.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "passport-strategy": "^1.0.0",
     "pg-migrator": "^1.0.5",
     "priorityqueuejs": "^1.0.0",
+    "promise-retry": "^1.1.1",
     "riverpig": "^1.0.0",
     "spdy": "^3.2.3",
     "tweetnacl": "^0.14.3",
--- a/test/fulfillmentSpec.js
+++ b/test/fulfillmentSpec.js
@@ -381,29 +381,39 @@ describe('PUT /fulfillment', function () {
       .expect(validator.validateTransfer)
       .end()
 
-    yield this.request()
-      .put(transfer.id + '/fulfillment')
-      .send(this.executionConditionFulfillment)
-      .expect(201)
-      .expect(this.executionConditionFulfillment)
-      .expect(validator.validateFulfillment)
-      .end()
+    const validateResponse = (res) => {
+      if (res.statusCode !== 200 && res.statusCode !== 201) {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -91,6 +91,7 @@
     "sinon-chai": "^2.7.0",
     "spec-xunit-file": "0.0.1-3",
     "supertest": "^1.1.0",
+    "supertest-as-promised": "^4.0.0",
```
justmoon: You don't need this. Just use `this.request()`...`.end()`  - That will return a promise.

test/fulfillmentSpec.js
```diff
@@ -348,6 +349,12 @@ describe('PUT /fulfillment', function () {
     })
 
   it('should not double spend when transfer is executed multiple times', function * () {
+    // Pre-fund hold account to allow double spending of source transfer
+    // This could happen if, for ex, 2 txn are escrowed, then first is executed twice.
```
justmoon: It seems like we should fix the first transaction being executed twice, then we wouldn't need this hack?

vhpoet: @justmoon @emschwartz I'm not very familiar with the code, does anyone want to look at this?

MatthewPhinney: Not sure what you mean by "transaction" (transfer? database transaction?). We had a separate issue we fixed with the notary, where it was sending two fulfillments per transfer. If we only get one fulfillment per transfer, this won't happen, but I think it's good to have a defensive test to confirm the ledger behaves properly, even in that case.

emschwartz:
Should we close this or update and merge it? @vhpoet @justmoon

justmoon:
@emschwartz Just went in and fixed the PR. Major changes, can you review please?

michielbdejong:
See also https://github.com/interledgerjs/five-bells-shared/pull/68 which already introduced retry logic at the `five-bells-shared` level

michielbdejong:
I'll try to merge the two as part of https://github.com/interledgerjs/five-bells-ledger/pull/408

PR #332, opened 2016-09-28
docs: explain CONNECTOR_PUBLIC_URI instead of deprecated options

justmoon:


.flowconfig
```diff
@@ -1,7 +0,0 @@
-[ignore]
-
-[include]
-
-[libs]
-
-[options]
```
emschwartz: What's this about?

justmoon: Belonged to Facebook Flow - no longer used.

PR #341, opened 2016-10-24
Maintenance and logging upgrade

justmoon:
Bunyan is causing issues during development due to the separate CLI transformer process. Riverpig is a cleaner, simpler logger. As a dependency it is 10x smaller than bunyan.

This PR also fixes a bunch of code rot:
- Update Node.js to 6
- Update dependencies
- Remove unintentional error messages emitted during successful websocket sends
- Various fixes related to dependency update


emschwartz:
LGTM except I would like to see more of an explanation of riverpig before we switch over to it

PR #348, opened 2016-10-31
feat: allow admin to send message as any user

### LINKED ISSUE #346: Allow admin to send message as any user
Admins should be able to impersonate any user. (This is used by the wallet/ilp-kit since the wallet/ilp-kit acts on behalf of its users.)

This means changing the `POST /messages` body to include a sender and receiver account and changing the authentication to work similar to the transfer authentication.

sentientwaffle:
Fixes #346

Depends on https://github.com/interledgerjs/five-bells-shared/pull/160

### DIFF
--- a/src/models/messages.js
+++ b/src/models/messages.js
@@ -14,8 +14,21 @@ function * sendMessage (message, requestingUser) {
     throw new InvalidBodyError(error, validationResult.errors)
   }
 
-  const recipientName = uri.parse(message.account, 'account').name.toLowerCase()
-  const senderAccount = uri.make('account', requestingUser.name)
+  const mode1 = !!message.account
+  const mode2 = !!(message.from && message.to)
+  if ((mode1 && mode2) || (!mode1 && !mode2)) {
+    throw new InvalidBodyError('Expected either message.account OR message.from/to')
+  }
+
+  const senderAccount = message.from || uri.make('account', requestingUser.name)
+  const senderName = uri.parse(senderAccount, 'account').name.toLowerCase()
+  const recipientName = uri.parse(message.to || message.account, 'account').name.toLowerCase()
+
+  // Only admin can impersonate users.
+  if (!requestingUser.is_admin && senderName !== requestingUser.name) {
+    throw new InvalidBodyError('You do not have permission to impersonate this user')
+  }
+
   yield notificationBroadcaster.sendMessage(recipientName, {
     ledger: message.ledger,
     account: senderAccount,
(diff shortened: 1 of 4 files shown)


test/postMessageSpec.js
```diff
@@ -109,4 +162,56 @@ describe('POST /messages', function () {
...
+      type: 'message',
+      resource: {
+        ledger: 'http://localhost',
+        account: 'http://localhost/accounts/alice',
+        data: {foo: 'bar'}
+      }
+    })
+  })
+
+  it('relays a message when the admin is impersonating another user', function * () {
```
justmoon: Should have a test to ensure regular users can't impersonate others.

src/models/messages.js
```diff
@@ -14,8 +14,21 @@ function * sendMessage (message, requestingUser) {
     throw new InvalidBodyError(error, validationResult.errors)
   }
 
-  const recipientName = uri.parse(message.account, 'account').name.toLowerCase()
-  const senderAccount = uri.make('account', requestingUser.name)
+  const mode1 = !!message.account
```
justmoon: I'd refactor this section to make it so that if `account` is set, the server:

* Asserts that `from`/`to` are not set
* Rewrites the object to use `from`/`to`
* Removes the `account` property

And then proceeds with the code as normal. Reason being that I really don't want to think of the two "modes" as equals. One is the deprecated way which is included for backwards compatibility and one is the normal way. So the compatibility code should be separate (and ideally have a comment stating "for backwards compatibility".)

test/postMessageSpec.js
sentientwaffle: It already does; see https://github.com/interledgerjs/five-bells-ledger/pull/348/files#diff-f4df84f46f3dcc8d94f3ede6fca6832eR124

src/models/messages.js
```diff
@@ -14,15 +14,16 @@ function * sendMessage (message, requestingUser) {
     throw new InvalidBodyError(error, validationResult.errors)
   }
 
-  const mode1 = !!message.account
-  const mode2 = !!(message.from && message.to)
-  if ((mode1 && mode2) || (!mode1 && !mode2)) {
-    throw new InvalidBodyError('Expected either message.account OR message.from/to')
+  // For backwards compatibility.
+  if (message.account) {
+    message.to = message.account
```
justmoon: You should still check (and throw an exception) if `to` or `from` are already (also) set.

justmoon: Nevermind, ignore that comment. Schema validation takes care of that.

PR #357, opened 2016-11-09
feat: BREAKING new websocket protocol

sentientwaffle:
Depends on https://github.com/interledgerjs/five-bells-shared/pull/162


test/notificationSpec.js
```diff
@@ -358,62 +372,241 @@ describe('Notifications', function () {
...
+      }))
+
+      // TODO: Is there a more elegant way?
+      yield timingHelper.sleep(50)
+
+      sinon.assert.calledOnce(listener)
+      sinon.assert.calledWithMatch(listener.firstCall, {
+        jsonrpc: '2.0',
+        id: 2,
+        error: { code: 403, message: 'Not authorized' }
```
emschwartz: I think it would be good if these errors also included the `id` field we send in the body for normal HTTP responses. The HTTP code is too general and the message isn't machine-readable

docs/apidoc_intro.ejs.md
```diff
@@ -112,3 +112,4 @@ Use the following environment variables to configure the service when run:
 * `LEDGER_AMOUNT_SCALE` (default: `2`) the number of digits allowed in amounts to the right of the decimal place
 * `LEDGER_LOG_LEVEL` (default: `info`) the allowed levels in order of verbosity are `fatal`, `error`, `warn`, `info`, `debug`, and `trace`
 * `LEDGER_RECOMMENDED_CONNECTORS` (default: `'*'`) a comma-delimited list of connector usernames
+* `LEDGER_AUTH_TOKEN_SECRET` (default: random bytes) a secret used to sign `/auth_token` tokens.
```
sentientwaffle: Do you mean a recommended length? The jwt is signed with HS256, so I don't think there's any strict length requirement being enforced.

src/controllers/accounts.js
```diff
@@ -111,87 +111,67 @@ function * putResource () {
...
  *    {
- *      "resource":{
- *        "debits":[
- *          {
- *            "account":"http://usd-ledger.exmaple/accounts/alice",
- *            "amount":"0.01",
- *            "authorized":true
+ *      "jsonrpc": "2.0",
+ *      "id": null,
+ *      "method": "notify",
```
sentientwaffle: Yes, it can either be a `Transfer` or a `Message`.

docs/apidoc_intro.ejs.md
emschwartz: Ok. I think @justmoon made a good point a while back that it would be best if the ledger just asked for one secret value and used that to generate all the others it needs

test/notificationSpec.js
sentientwaffle: The error object is specified by http://www.jsonrpc.org/specification#error_object. We could use `error.message` for the error `id`, and `error.data` for the error message, though.

emschwartz: I would be a fan of putting our normal error object in the data field for consistency. What would go in the message field though I'm not sure

emschwartz:
Also, @sentientwaffle please squash the merge commit before merging into master

docs/apidoc_intro.ejs.md
```diff
@@ -112,3 +113,21 @@ Use the following environment variables to configure the service when run:
 * `LEDGER_AMOUNT_SCALE` (default: `2`) the number of digits allowed in amounts to the right of the decimal place
 * `LEDGER_LOG_LEVEL` (default: `info`) the allowed levels in order of verbosity are `fatal`, `error`, `warn`, `info`, `debug`, and `trace`
 * `LEDGER_RECOMMENDED_CONNECTORS` (default: `'*'`) a comma-delimited list of connector usernames
+* `LEDGER_SECRET` (default: random bytes) a secret used to sign `/auth_token` tokens. Any length.
+
+## RPC Error Codes
+<a id='rpc_error_codes'></a>
+
+For errors which have an equivalent in HTTP, we choose a five digit error code where the first three digits correspond to the HTTP status code.
```
sentientwaffle: Yes, in the websocket channel. The negative codes are specified in http://www.jsonrpc.org/specification.

src/lib/rpcHandler.js
```diff
@@ -0,0 +1,136 @@
...
+
+      if (reqMessage.method === 'subscribe_account') {
+        resMessage.result = this.subscribeAccount(reqMessage.params.eventType,
+          reqMessage.params.accounts)
+      } else {
+        throw new RpcError(-32601, 'Unknown method: ' + reqMessage.method)
+      }
+    } catch (err) {
+      resMessage.error = {
+        code: err instanceof SyntaxError ? -32700 : (err.code || 50000),
```
emschwartz: Instead of having these codes scattered throughout the code it would be better to have one file that's required in a number of places that would export constants like `websocketErrorCodes.SYNTAX_ERROR`

sentientwaffle: Should the tests use the raw values, or reference the constants?

emschwartz: I don't have a strong opinion about that

PR #369, opened 2016-11-22
feat: NoSubscriptionsError, verify connectors

sentientwaffle:
* Warn on startup if a recommended connector doesn't exist.
* Return a `NoSubscriptionsError` error on `POST /messages` if the receiving account has no active websockets.

Fixes https://github.com/interledgerjs/five-bells-ledger/issues/362


justmoon:
LGTM

PR #373, opened 2016-11-28
fix: set ws.Server maxPayload to prevent oom

sentientwaffle:


src/lib/app.js
```diff
@@ -55,6 +56,16 @@ class App {
     koaApp.ws.use(passport.initialize())
     koaApp.ws.use(websocketRouter.routes())
     koaApp.ws.use(websocketRouter.allowedMethods())
+    // Override listen() in order to specify a maxPayload. See:
+    //   https://github.com/kudos/koa-websocket/blob/785001c20d94fbb60cb05596bb1869889f539351/index.js#L15-L20
+    // for the original version.
+    koaApp.ws.listen = function (server) {
```
emschwartz: If you need to overwrite one of 3 methods provided by this library is it even worth using the library? If we do stick with this we should submit a PR or file an issue on the module's repo about allowing you to specify additional options.

src/lib/app.js
```diff
@@ -55,6 +56,16 @@ class App {
...
+      this.server = new WebSocketServer({
+        server: server,
+        maxPayload: 64 * 1024
```
emschwartz: It would be good to put this as a constant up at the top of the file and provide a justification for the size chosen

sentientwaffle:
@emschwartz ok, fixed

src/lib/koa-websocket.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
```
sentientwaffle: Ok. There's already a month-old issue at https://github.com/kudos/koa-websocket/issues/22.

emschwartz: Gotcha. In that case, sounds fine to copy the code locally and remove the dependency on that package

PR #375, opened 2016-11-29
fix: limit HTTP request body size

sentientwaffle:


dappelt:
The parse function of `co-body` is also used in [src/controllers/transfers.js](https://github.com/interledgerjs/five-bells-ledger/blob/master/src/controllers/transfers.js).

Is setting the size limit in `app.js` sufficient to also enforce the size limit in said file?

PR #379, opened 2016-12-01
fix: use correct id format, schema name

sentientwaffle:


src/controllers/transfers.js
```diff
@@ -293,7 +293,7 @@ function * putResource () {
     )
   }
 
-  transfer.id = id.toLowerCase()
+  transfer.id = uri.make('transfer', id.toLowerCase())
```
emschwartz: Shouldn't there be a test for this?

sentientwaffle: The old version didn't cause any errors, but with https://github.com/interledgerjs/five-bells-shared/pull/169 it does because the UUID no longer matches the URI schema.

PR #389, opened 2016-12-09
feat: add back support for global subscriptions

justmoon:


test/notificationSpec.js
```diff
@@ -767,6 +767,79 @@ describe('Notifications', function () {
     })
   })
 
+  describe('GET /websocket method:subscribe_all_accounts as admin', function () {
```
emschwartz: It would be good to test that it doesn't allow non-admins to subscribe

momerath42 (approved):
works for me

PR #391, opened 2016-12-13
fix: fix misc. logging issues

sentientwaffle:
* Don't use `logger` for websocket connections, it tries to attach events to `this.res`, which doesn't exist.
* More detailed logging when pg-migrator fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "koa-cors": "0.0.16",
     "koa-mag": "^1.0.4",
     "koa-passport": "^1.1.6",
-    "koa-riverpig": "^1.0.0",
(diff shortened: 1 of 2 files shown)


src/lib/app.js
```diff
@@ -50,7 +50,6 @@ class App {
     koaApp.use(compress())
 
     const websocketRouter = this._makeWebsocketRouter()
-    koaApp.ws.use(logger)
```
emschwartz: Is there a way to make it log activity in another way? Or do we just not want it to log activity?

sentientwaffle: Fixed. The issue is fixed in the new version of koa-riverpig, so I can leave the logging in.

PR #394, opened 2017-01-04
feat: add send_message rpc method

sentientwaffle:
Depends on https://github.com/interledgerjs/five-bells-shared/pull/176


sentientwaffle:
Closing without merge to avoid breaking ledger changes.

PR #400, opened 2017-02-22
perf: Change default scale,precision to 10 digits before the comma, 9 after

michielbdejong:
if the currency is for instance dollars, precision will now be nanodollars, and maximum balance will
be 10 billion dollars

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -17,8 +17,8 @@ function useTestConfig () {
 
 function parseAmountConfig () {
   return {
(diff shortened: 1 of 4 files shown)


justmoon:
New scale fully includes the old scale, so it's technically not breaking anything, it may just be surprising to you that your ledger suddenly accepts higher precision transfers. I'd say minor version should be ok.

michielbdejong:
Can't check what circleci is complaining about because circleci.com is down

PR #403, opened 2017-03-15
fix: fixes admin account setup for postgres

dappelt:
- Fixes the setup of the admin account
- Adds a check that minimum_allowed_balance is numeric when creating new accounts


dappelt:
I found an easier solution. Closing this PR in favor of PR #404

PR #404, opened 2017-03-16
fix: fixes admin account setup

dappelt:
- fixes setup of the admin account if postgres is used
- adds a unit test to check if the admin account was correctly setup

### DIFF
--- a/src/lib/seed-db.js
+++ b/src/lib/seed-db.js
@@ -2,6 +2,7 @@
 const hashPassword = require('five-bells-shared/utils/hashPassword')
 const getAccount = require('../models/db/accounts').getAccount
 const upsertAccount = require('../models/db/accounts').upsertAccount
+const insertAccounts = require('../models/accounts').insertAccounts
 
 module.exports = function * (config) {
   yield setupHoldAccount()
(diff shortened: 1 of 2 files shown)


src/lib/seed-db.js
```diff
@@ -29,13 +30,13 @@ function * setupAdminAccount (adminParams) {
     adminAccount.fingerprint = adminParams.fingerprint
     yield upsertAccount(adminAccount)
   } else {
-    yield upsertAccount({
+    yield upsertAccount(converters.convertToInternalAccount({
```
sentientwaffle: Wouldn't it be easier just to pass an internal account, since its being constructed here anyway? Whichever route you go, the `upsertAccount()` call in `setupHoldAccount()` needs the same treatment.

dappelt: I changed this line to use [insertAccounts()](https://github.com/interledgerjs/five-bells-ledger/blob/master/src/models/accounts.js#L98).

I think it is more robust to pass in an external account to take advantage of all the input validation that is done during the conversion from external to internal account. In the future, [small changes to the admin account setup](https://github.com/interledgerjs/five-bells-ledger/commit/e994f2f7c7713bc2ce8fb3acaa20348c577e93e2) are less likely to result in failing database queries.

sentientwaffle:
Try rebasing onto master and see if that makes the tests pass.

PR #407, opened 2017-03-21
readme fixes and sqlite3 devDependency

michielbdejong:
sqlite3 is needed for 'Step 3' from the readme, as well as for `npm test`.

The Docker-based instructions still referred to mysql, updated it to postgres.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,26 +61,31 @@ Open `apidocs-out/index.html` in a web browser to see the generated API document
 
 This project can be run in a [Docker](https://www.docker.com/) container.
 
-You need a local database instance listening on port 8080. Here is how to set that up:
+You need to start a postgres container:
 
 ``` sh
-docker run --name mysql -e MYSQL_ROOT_PASSWORD=password -p 3306:3306 -d mysql
-export LEDGER_DB_URI=mysql://root:password@localhost/fivebells
-npm run migrate
+docker run --name five-bells-ledger-db -e POSTGRES_PASSWORD=password -d postgres
 ```
 
-Then run the following (with the same environment variables) as described above:
+After giving postgres a few seconds to start up, you can run a five-bells-ledger Docker container, linking to that database:
 
 ``` sh
-docker run -it --rm --net=host -e LEDGER_PORT=1337 -e LEDGER_DB_URI=$LEDGER_DB_URI interledger/five-bells-ledger
+docker run -it --rm -e LEDGER_PORT=1337 -e LEDGER_DB_URI=postgres://postgres:password@db --link five-bells-ledger-db:db -p 1337:1337 -h localhost --name fivebells interledger/five-bells-ledger
 ```
 
 Breaking down that command:
 
 * `-it` Run Five Bells Ledger in an interactive terminal.
 * `--rm` Delete container when it's done running.
-* `--net=host` Don't isolate container into its own virtual network. This allows Five Bells Ledger to see the database that we set up above.
 * `-e LEDGER_PORT=1337` Set the ledger's port to 1337. This is just an example for how to set a config option.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -64,23 +64,32 @@ This project can be run in a [Docker](https://www.docker.com/) container.
 You need a local database instance listening on port 8080. Here is how to set that up:
 
 ``` sh
-docker run --name mysql -e MYSQL_ROOT_PASSWORD=password -p 3306:3306 -d mysql
-export LEDGER_DB_URI=mysql://root:password@localhost/fivebells
-npm run migrate
+docker run --name db -e POSTGRES_PASSWORD=password -p 5432:5432 -d postgres
```
justmoon: Would prefer a name that's more descriptive and less likely to collide with a container the user already has, e.g. `--name five-bells-ledger-db`.

README.md
```diff
@@ -64,23 +64,32 @@ This project can be run in a [Docker](https://www.docker.com/) container.
...
+export LEDGER_DB_URI=postgres://postgres:password@localhost
+docker run -it --rm --link db:postgres postgres psql -h postgres -U postgres
```
justmoon: Remove this line.

README.md
```diff
@@ -64,23 +64,32 @@ This project can be run in a [Docker](https://www.docker.com/) container.
 You need a local database instance listening on port 8080. Here is how to set that up:
 
 ``` sh
-docker run --name mysql -e MYSQL_ROOT_PASSWORD=password -p 3306:3306 -d mysql
-export LEDGER_DB_URI=mysql://root:password@localhost/fivebells
-npm run migrate
+docker run --name db -e POSTGRES_PASSWORD=password -p 5432:5432 -d postgres
+export LEDGER_DB_URI=postgres://postgres:password@localhost
```
justmoon: I'd prefer if we got rid of the `-p 5432:5432`. Again, because it's likely to collide with another postgres server.

Instead, you can do:

``` sh
export LEDGER_DB_URI=postgres://postgres:password@`docker inspect --format '{{ .NetworkSettings.IPAddress }}' five-bells-ledger-db`
```

README.md
```diff
@@ -64,23 +64,32 @@ This project can be run in a [Docker](https://www.docker.com/) container.
...
+docker run -it --rm --link db:postgres postgres psql -h postgres -U postgres
+docker run -it --rm --link db:postgres -v `pwd`:/src postgres psql -h postgres -U postgres
+	postgres=# \i /src/src/sql/pg/1-2.sql
```
justmoon: Total nitpick, but `/src/src` is a bit confusing, so why not mount it under `/var/app`, `/opt/five-bells-ledger` or something like that.

README.md
michielbdejong: good point! fixed

README.md
michielbdejong: good point as well, and actually turns out it's not needed to expose the postgres db port to the host because we use container linking

README.md
michielbdejong: yes, good catch

README.md
michielbdejong: I now made it mount `src/` as `/src/`, which also avoids the ugly `/src/src` path.

I also changed the psql cli arguments around a bit so no longer needs to be interactive

michielbdejong:
@justmoon thanks for helpful review! Change requests addressed.

michielbdejong:
will do sqlite3 dependency separately

michielbdejong:
@dappelt can you review the removal of sqlite3 from devDependencies? I explained the reasoning in [the commit message](https://docs.npmjs.com/files/package.json#optionaldependencies)

PR #408, opened 2017-03-23
fix: retry fulfillTransfer when hold account is busy

michielbdejong:


src/models/transfers.js
```diff
@@ -435,7 +435,8 @@ function * fulfillTransfer (transferId, fulfillmentUri) {
       }
     } catch (err) {
       // In case of a synchronization error, retry
-      if (Number(err.code) === 40001) {
+      if (Number(err.code) === 40001 || err.name === 'InsufficientFundsError') {
+        log.debug('hold account busy, retrying fulfill');
```
michielbdejong: I'll remove the `;`, sorry

test/helpers/dbFailureMock.js
```diff
@@ -0,0 +1,22 @@
+var mock = require('mock-require');
```
michielbdejong: I'll use `const`, sorry

test/helpers/dbFailureMock.js
```diff
@@ -0,0 +1,22 @@
...
+
+var RealUtils = require('../../src/models/db/utils');
+var mockUtils = function() {
+  var ret1 = RealUtils.apply(null, arguments);
```
michielbdejong: I'll use more descriptive variable names than `ret1` and `ret2`

michielbdejong:
@emschwartz you can now see the retry in action using:

```sh
SHOW_STDOUT=true DEBUG=ledger:transfers npm test
```
I'm still looking into how to make the tests run faster, but otherwise this patch should hopefully be pretty close.

src/models/transfers.js
```diff
@@ -435,13 +437,18 @@ function * fulfillTransfer (transferId, fulfillmentUri) {
       }
     } catch (err) {
       // In case of a synchronization error, retry
-      if (Number(err.code) === 40001) {
-        retry(err)
+      if (Number(err.code) === 40001 || err.name === 'InsufficientFundsError') {
```
dappelt: Instead of always retrying whenever there is an `InsufficientFundsError`, should we better differentiate what the reason for the error was and only retry when the database transaction fails because it cannot be serialized?

emschwartz: See https://github.com/interledgerjs/five-bells-shared/blob/master/lib/db.js#L84-L87

michielbdejong: @dappelt yeah, good point, I'll fix https://github.com/interledgerjs/five-bells-ledger/blob/2c5d1abc7112c5fac6e111d2b9a0fd42e8471278/src/lib/holds.js#L10-L13 so that it lets through the original error.

@emschwartz oh wow, seems we're doing two layers of retry, then, [once in five-bells-shared/lib.db.js](https://github.com/interledgerjs/five-bells-shared/blob/master/lib/db.js#L72) using the [retry package](https://www.npmjs.com/package/retry), and [once in five-bells-ledger/src/model/transfers.js](https://github.com/interledgerjs/five-bells-ledger/blob/master/src/models/transfers.js#L381) using the [promise-retry package](https://www.npmjs.com/package/promise-retry). Synchronization error 40001 seems to be specific to IBM DB2. Do you want to

emschwartz: I don't think there should be two retry layers

michielbdejong: ok, I'll dig into that

michielbdejong: turns out that `lib/db.js` from five-bells-ledger is no longer used by five-bells-ledger, opened an [issue](https://github.com/interledgerjs/five-bells-shared/issues/186) about that, and so there are no two retry layers. I'll still change which error comes through from src/lib/holds.js, though.

michielbdejong:
I'll change this PR so that it retries a the `withSerializeableTransaction` level

src/lib/holds.js
```diff
@@ -8,6 +8,9 @@ function adjustBalance (accountName, amount, transaction) {
   /* eslint-disable handle-callback-err */
   return accounts.adjustBalance(accountName, amount, {transaction})
   .catch((error) => {
+    if (error.code === '40001') {
```
emschwartz: Please include a comment describing what this error means

test/fulfillmentSpec.js
```diff
@@ -177,6 +178,60 @@ describe('GET /fulfillment', function () {
...
+
+    afterEach(function () {
+      const timesLeft = dbFailureMock.timesQueryShouldFail[JSON.stringify(ARGS_EXPECTED)]
+      dbFailureMock.timesQueryShouldFail[JSON.stringify(ARGS_EXPECTED)] = 0
+      expect(timesLeft).to.equal(0)
+    })
+
+    it('should return a fulfillment', returnFulfillment)
+
+    describe('for a very long time', function () {
```
emschwartz: Since there is only one test under this `describe`, I think it would be better to just have the test by itself and to just put the setup / teardown steps in the test (especially because the after includes an assertion that I think it actually part of the test)

src/models/transfers.js
```diff
@@ -49,6 +48,8 @@ const RECEIPT_TYPE_SHA256 = 'sha256'
 const CONDITION_TYPE_EXECUTION = 'execution'
 const CONDITION_TYPE_CANCELLATION = 'cancellation'
 
+const MAX_RETRY = 100
```
dappelt: I can't find where this variable is used.

src/lib/db.js
```diff
@@ -26,22 +27,35 @@ const TABLE_NAMES = [
   'L_LU_TRANSFER_STATUS'
 ]
 
+const MAX_RETRY = 1000
```
dappelt: Is 1000 retries a reasonable number? Would it make sense to let transactions fail after fewer attempts?

I wonder if a high number of retries can be counterproductive. For example, failing retries from different transactions might stack up and eventually slow down the database.

michielbdejong: right, and maybe you would want the number of retries be higher for fulfillments than for initiating new transactions. I'll continue working on this next week.

src/models/transfers.js
michielbdejong: oops, forgot to remove that, thanks.

michielbdejong:
This PR adds retries to the `withSerializableTransaction` method, and in doing so, fixes the retries for fulfilling a transfer, and adds tests for that.
It also adds retries for updateTransfer and upsertTransfer (through upsertAdjustments), but there are no tests for that at all, so I also didn't update any tests there.
It also adds retries to expireTransfer, which is used from the timerWorker, will see if I can adjust the timerWorkerSpec for this change.

michielbdejong:
It also affects putResource, through model.setTransfer. I'll see if i can add a retry test there.

michielbdejong:
apart from adding more tests, other review comments all addressed.

src/lib/db.js
```diff
@@ -26,22 +28,40 @@ const TABLE_NAMES = [
...
+      // 40001 is a postgres error code meaning the database could not complete the transaction
+      // because this would interfere with other concurrent transactions
+      if (err.code === '40001') {
+        log.debug('retrying database query', `${attemptNo}/${retries}`, err)
+        retry(err)
+      }
+      throw err
+    })
+  }, {
+    minTimeout: 0,
```
dappelt: By setting `minTimeout: 0` you retry immediately after the transaction fails. Is this intended? Would it make sense waiting for a little while before retrying?

For fulfillments I can see why you want to do it as fast as possible, but what about other transactions?

michielbdejong: ok, my assumption was that the timescale of database commits is way smaller than that of a query roundtrip, but maybe that's not true in practice. Will set it to 1000 microseconds.

src/lib/transferExpiryMonitor.js
```diff
@@ -54,7 +55,7 @@ class TransferExpiryMonitor {
 
         yield _this.notificationBroadcaster.sendNotifications(transfer, transaction)
       }
-    })
+    }, 10) // retries if database is busy
```
emschwartz: Why 10 times? Constants like this should be declared at the top of the file so they're easier to find later

src/models/db/adjustments.js
```diff
@@ -101,8 +101,9 @@ function upsertAdjustment (persistentAdjustment, options) {
   if (options && options.transaction) {
     return _upsertAdjustment(persistentAdjustment, options.transaction)
   } else {
+    // second parameter is number of retries in case database is busy
     return db.withSerializableTransaction((transaction) =>
-      _upsertAdjustment(persistentAdjustment, transaction))
+      _upsertAdjustment(persistentAdjustment, transaction), 5)
```
emschwartz: Why 5? Same as above, this kind of constant should be declared at the top of the file

src/models/transfers.js
```diff
@@ -380,70 +378,58 @@ function * fulfillTransfer (transferId, fulfillmentUri) {
...
+    yield db.updateTransfer(transfer, {transaction})
+
+    // Start the expiry countdown if the transfer is not yet finalized
+    // If the expires_at has passed by this time we'll consider
+    // the transfer to have made it in before the deadline
+    if (!isTransferFinalized(transfer)) {
+      yield transferExpiryMonitor.watch(transfer)
     }
-  }))
+  }, 50) // retry fulfillments this many times if db  is busy
```
emschwartz: Declare at top

test/fulfillmentSpec.js
```diff
@@ -177,6 +178,57 @@ describe('GET /fulfillment', function () {
...
+      try {
+        const transfer = this.preparedTransfer
+
+        yield dbHelper.setHoldBalance(10)
+
+        yield this.request()
+          .put(transfer.id + '/fulfillment')
+          .auth('admin', 'admin')
+          .send(this.executionConditionFulfillment)
+          .expect(500)
```
emschwartz: This should probably return a `503 Service Unavailable` error. Generally you don't expect things to throw 500 errors

test/notificationSpec.js
```diff
@@ -13,6 +13,8 @@ const transferExpiryMonitor = require('../src/services/transferExpiryMonitor')
 const transferDictionary = require('five-bells-shared').TransferStateDictionary
 const transferStates = transferDictionary.transferStates
 
+const dbFailureMock = require('../src/models/db/utils')
```
emschwartz: Isn't this in `test/helpers/`?

michielbdejong: '../src/models/db/utils' was mock-required by the helper. i'll add a comment, though.

test/fulfillmentSpec.js
michielbdejong: done. I added the 503 responses in all API points for PUT, not for the GET ones.
Do you also want me to create unit tests that check the 503 responses for the other API end points, or is it ok like this?

michielbdejong:
Needs a rebase because of https://github.com/interledgerjs/five-bells-ledger/pull/417

test/fulfillmentSpec.js
michielbdejong: @emschwartz if the answer was yes (or if there's anything else that you still want to see improved), then please let me know, and I'll create a follow-up ticket. Merging this now, with all your comments addressed, hopefully.

michielbdejong:
@dappelt I think I addressed all your review comments now, as well as @emschwartz's - can you give this a PR a final review check?

Hope the rebase didn't introduce new issues. At least the tests are all passing now.

src/lib/db.js
```diff
@@ -25,20 +27,37 @@ const TABLE_NAMES = [
...
+        log.debug('retrying database query', `${attemptNo}/${retries}`, err)
+        err.isDbRetry = true
+        return retry(err)
+      }
+      throw err
+    })
+  }, {
+    minTimeout: 1, // 1000 microseconds
+    factor: 1,
+    retries
```
dappelt: I did some performance testing and found these retry settings are too aggressive. I propose the following settings:
```
minTimeout: 10, // leave more time between retries
factor: 1.2, // consecutive retries have an increasing timeout
retries,
randomize: true // randomizing the duration after which the retry occurs helps if a batch of transfers/fulfillments are submitted at the same time. Since the retries do not happen all at the same time, they have a higher likelihood of succeeding.
```

michielbdejong: done. thanks!

PR #409, opened 2017-03-24
docs: Add more usage info to the readme

michielbdejong:
The env var for setting the admin user password was missing, so you couldn't do anything with the ledger from the Docker instructions.

These additions also explain how to create a user, how to inspect the database, how to connect a plugin, and add a word of warning about reliability.


michielbdejong:
what is `Error: Cannot find module 'five-bells-service-manager'`? Does that come from the integration tests? Pretty sure I didn't break that with these readme changes ;)

dappelt:
@michielbdejong Try rebuilding without cache, should fix it.

PR #410, opened 2017-03-24
feat: Announce API version in metadata

michielbdejong:
There is currently no way to know which version a ledger is running, other than if the ledger was discovered through an ilp-kit WebFinger record. Adding this, fixes that.


michielbdejong:
Would not every API change cause the five-bells-ledger major version to change?
And what could cause the five-bells-ledger major version to change, if not an API change?

emschwartz:
Fair point. The only two things I can think of that might cause a major version change without an API change would be something like changing configuration options or removing a supported database.

emschwartz:
Maybe just `version` would be a more appropriate name

PR #412, opened 2017-03-27
feat: checks max. once per minute if the DB is still alive

dappelt:
It should not be possible to create a non-neglectable workload via unauthenticated endpoints such as /health. This commit limits the number of database queries that are trigger by requesting /health to 1 per minute.

### DIFF
--- a/src/controllers/health.js
+++ b/src/controllers/health.js
@@ -1,7 +1,10 @@
 'use strict'
 
 const model = require('../models/health')
+const cacheExpiry = 60000
 
+let lastQueried
+let cachedResult
 /**
  * @api {get} /health Get server health status
  * @apiName GetHealth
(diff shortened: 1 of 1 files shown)


src/controllers/health.js
```diff
@@ -17,5 +18,11 @@ const model = require('../models/health')
  * @returns {void}
  */
 exports.getResource = function * health () {
-  this.body = yield model.getDbHealth()
+  const now = Date.now()
+  // call model.getDbHealth(...) max. once per minute
+  if (!lastQueried || lastQueried > (now - 60000)) {
```
emschwartz: Constants like the `60000` should be declared at the top of the file as `const`'s

emschwartz:
This was replaced by https://github.com/interledgerjs/five-bells-ledger/pull/413, no?

PR #414, opened 2017-03-28
feat: changes unauthenticated requests to getAccount() to not hit the db

dappelt:


src/models/accounts.js
```diff
@@ -44,19 +44,25 @@ function * getAccounts () {
...
+      throw new NotFoundError('Unknown account')
+    } else if (account.is_disabled &&
       (requestingUser && !requestingUser.is_admin)) {
-    throw new UnauthorizedError('This account is disabled')
+      throw new UnauthorizedError('This account is disabled')
+    }
+
+    // TODO get rid of this when we start using biginteger math everywhere
+    account.balance = Number(account.balance).toString()
+    delete account.password_hash
```
michielbdejong: wouldn't it be safer to define which fields to include instead of which ones to exclude?

src/models/accounts.js
```diff
@@ -44,19 +44,25 @@ function * getAccounts () {
...
-  const canExamine = requestingUser &&
-    (requestingUser.name === name || requestingUser.is_admin)
-  const account = yield db.getAccount(name)
-  if (!account) {
-    throw new NotFoundError('Unknown account')
-  } else if (account.is_disabled &&
+  let canExamine = false
+  let account
+  if (!requestingUser) { // anonymous request
+    account = {id: name, name: name}
```
michielbdejong: shouldn't id be a URL? see:

```
../../interledgerjs/five-bells-ledger/test//data/accounts.json:  "candice": {
../../interledgerjs/five-bells-ledger/test//data/accounts.json:    "id": "http://localhost/accounts/candice",
../../interledgerjs/five-bells-ledger/test//data/accounts.json:    "name": "candice",
...
```

dappelt: This object is not the object that is returned in the HTTP response. Once it has [been converted](https://github.com/interledgerjs/five-bells-ledger/blob/f4055ea73bad19211c96a96d0619acce93fc4d66/src/models/accounts.js#L67), `id` is an URL.

src/models/accounts.js
dappelt: I agree although I did not add this line. It is already in master. 

In addition, it seems to be redundant since `converter.convertToExternalAccount(...)` in [line 66](https://github.com/interledgerjs/five-bells-ledger/blob/f4055ea73bad19211c96a96d0619acce93fc4d66/src/models/accounts.js#L66) remove this property too.

Does anybody see a reason why we should not drop this line?

michielbdejong:
@dappelt can this be merged?

michielbdejong:
@dappelt anything I can do to help speed this up?

dappelt:
This PR does not fully mitigate the problem I set out to fix, so I might have to tweak it a little bit. Will keep you posted.

PR #417, opened 2017-04-12
Update dependencies (specifically, update to Koa 2 + async/await)

justmoon:
Big update in anticipation of the impending release of Node 8.

* Updated dependencies
* generators -> async/await
* Removed co
* Added a `yarn.lock` - doesn't hurt to have it.
* co-supertest -> supertest: required removing the `end()` calls. If you keep them, it silently ignores all assertions, not good. 
* Kept co-body (there's koa-bodyparser, but that just uses co-body internally, so...)
* Removed co-mocha (mocha natively supports promises, so we can just use async functions now)


src/controllers/metadata.js
```diff
@@ -75,14 +76,14 @@ module.exports = (config) => {
   }
 
   try {
-    metadata.version = `five-bells@${require('../../package.json').version.split('.')[0]}`
```
justmoon: This was pulled out to the top of the file, because having a require in a string template made `istanbul` (code coverage tool) choke.

PR #419, opened 2017-05-05
feat: logs the error message if DB tables cannot be created/migrated

dappelt:


dappelt:
`createTable()` does not throw if the DB already exists

sharafian:
If I run a ledger for the first time with a database, I get:

```
2017-05-05T14:29:41.529Z ledger:app info syncing database
2017-05-05T14:29:41.620Z ledger:app info ledger listening on 127.0.0.1:3000
2017-05-05T14:29:41.621Z ledger:app info public at http://localhost:3000
```

If I run it again on the same database, I get:

```
2017-05-05T14:30:24.383Z ledger:app info syncing database
{ Error: SQLITE_ERROR: index accounts_name_unique already exists errno: 1, code: 'SQLITE_ERROR' }
2017-05-05T14:30:24.405Z ledger:app info database sync aborted
2017-05-05T14:30:24.435Z ledger:app info ledger listening on 127.0.0.1:3000
2017-05-05T14:30:24.435Z ledger:app info public at http://localhost:3000
```

If the index exists already it throws an error, but it shouldn't be treated as an unexpected error and printed with a stack trace

dappelt:
That's strange. For me it is not throwing if the DB exists. What version of postgres do you use?

dappelt:
Oh, I see .. it's SQLite ...

sharafian:
Well, if sqlite is only being used for development, maybe it's ok to have an error message printed. On the other hand, it might throw people off while debugging.

Maybe we should decide whether to stop supporting anything other than postgres in the ledger. You mentioned that there were a lot of useful postgres specific features that we could be making use of.

dappelt:
I added code that checks if it is an error due to the DB is already existing.

src/lib/app.js
```diff
@@ -71,7 +71,10 @@ class App {
       this.log.info('syncing database')
       await createTables()
     } catch (e) {
-      this.log.info('database sync aborted')
+      const dbExistsError = e.message && e.message.includes('already exists')
```
sharafian: We should probably be more specific about this message

PR #422, opened 2017-05-12
perf: remove hold account

justmoon:
The hold account no longer serves any useful purpose, but it does create a major bottleneck, because the database has to constantly lock the row containing the hold account.


michielbdejong:
Good idea! This means the sum over all account balances will no longer be zero, but I guess that's OK. If someone wants to know the balance of the hold account, they can derive it that way. :)

michielbdejong:
merged in #426.

PR #425, opened 2017-05-29
feat: adds token authentication to all endpoints

dappelt:


test/putTransferSpec.js
```diff
@@ -571,6 +571,15 @@ describe('PUT /transfers/:id', function () {
...
 
+  it('should return 403 if invalid auth token is given', async function () {
+    const transfer = this.transferFromEve
+
+    await this.request()
+      .put(transfer.id + '?token=wrong')
+      .send(transfer)
+      .expect(403)
+  })
+
```
emschwartz: Also add a test to make sure a token from another user doesn't work

src/lib/tokenStrategy.js
```diff
@@ -10,10 +10,10 @@ class TokenStrategy extends passport.Strategy {
 
   authenticate (req) {
     const token = req.query.token
```
emschwartz: As discussed, this should be switched from putting the token in the query string to adding a bearer token auth header

PR #427, opened 2017-06-08
fixed docker run command - password env variable

andreasp1994:
LEDGER_ADMIN_PASS is the env variable that should be used. It took me a while to figure it out while trying to run it from docker. This should save some users from unnecessary problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ docker run --name five-bells-ledger-db -e POSTGRES_PASSWORD=password -d postgres
 After giving postgres a few seconds to start up, you can run a five-bells-ledger Docker container, linking to that database:
 
 ``` sh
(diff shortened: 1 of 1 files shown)


michielbdejong:
Oh, bummer! Looks like none of the maintainers saw this PR, because that was an important fix, and I just merged exactly the same fix in #432. I know for myself I accidentally wasn't watching this repo, I switched to 'Watching' last week. I'll contact the rest of the team members from https://github.com/orgs/interledgerjs/teams/interledger-js-maintainers/members to make sure enough people are watching each and every repo under the org, and we don't let this happen again.

 Anyway, @andreasp1994 thanks for the fix, it's merged as part ofhttps://github.com/interledgerjs/five-bells-ledger/commit/4b422ac64df302ee2e51014ef0f1745733001d97!

PR #428, opened 2017-06-12
fix typo in circle-ci script

michielbdejong:


michielbdejong:
This change is not actually tested by circleci because that uses 4 threads, but it seems like an obvious typo to me.

PR #429, opened 2017-06-12
feat: add token_max_age to /auth_token

sentientwaffle:
Related to https://github.com/interledgerjs/ilp-plugin-bells/pull/136#issuecomment-307850559

### DIFF
--- a/src/controllers/authTokens.js
+++ b/src/controllers/authTokens.js
@@ -29,7 +29,7 @@ const config = require('../services/config')
 async function getAuthToken (ctx) {
   ctx.body = {
     token: await model.getAuthToken(ctx.state.user),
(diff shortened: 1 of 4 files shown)


src/controllers/authTokens.js
```diff
@@ -18,13 +19,17 @@ const model = require('../models/authTokens')
  *    HTTP/1.1 200 OK
  *    Content-Type: application/json
  *
- *    { "token": "9AtVZPN3t49Kx07stO813UHXv6pcES" }
+ *    {
+ *      "token": "9AtVZPN3t49Kx07stO813UHXv6pcES",
+ *      "token_max_age": 604800000
```
emschwartz: It would be good to include a note somewhere about the units of this field

PR #436, opened 2017-11-22
Allow passing back of fulfillment data

justmoon:
See interledger/rfcs#338


src/controllers/transfers.js
```diff
@@ -343,7 +402,7 @@ async function putFulfillment (ctx) {
 async function getFulfillment (ctx) {
   const id = ctx.params.id
   requestUtil.validateUriParameter('id', id, 'Uuid')
-  ctx.body = await model.getFulfillment(id.toLowerCase())
+  ctx.body = (await model.getFulfillment(id.toLowerCase())).condition_fulfillment
```
sharafian: this should return the fulfillment data too, so retries don't just get part of the info back

src/sql/pg/2-3.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "L_FULFILLMENTS"
+  ADD COLUMN "FULFILLMENT_DATA" CHARACTER VARYING(4000) NULL;
```
sharafian: this should be 32k for the max ilp packet size

src/controllers/transfers.js
justmoon: Added a `GET /fulfillment2` endpoint that returns both.

